PR #14, opened 2015-12-03
Add generateDockerCompose task and 'docker' component for publishing …

uschi2000:
…docker image dependencies

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -33,10 +33,110 @@ Configuration specifying all parameters:
         dockerfile 'Dockerfile'
         dependsOn tasks.distTar
         files 'file1.txt', 'file2.txt'
+        dockerComposeTemplate 'my-template.yml'
+        dockerComposeFile 'my-docker-compose.yml'
     }
 
 To build a docker container, run the `docker` task. To push that container to a
 docker repository, run the `dockerPush` task.
+
+
+Generating docker-compose dependencies
+--------------------------------------
+
+The plugin provides mechanisms for managing dependencies between docker images
+used in orchestrating `docker-compose` environments. Each project can declare
+its dependencies on other docker images and publish an artifact advertising those
+dependencies. The plugin uses standard Maven/Ivy machanism for declaring and
+resolving dependencies.
+
+The `generateDockerCompose` task generates a `docker-compose.yml` file from
+a user-defined template by replacing each version variable by the concrete version
+declared by the transitive dependencies of the docker configuration.
+
+### Specifying dependencies on Docker images
+The plugin adds a `docker` Gradle component and a `docker` Gradle configuration that
+can be used to specify and publish dependencies on other Docker containers.
+The `generateDockerCompose` task (see below) uses the transitive dependencies
+of the `docker` configuration to populate a `docker-compose.yml.template` file
+with the image versions specified by this project and all its transitive
+dependencies.
+
+    dependencies {
+        docker 'foogroup:barmodule:0.1.2'
+        docker project(":someSubProject")
+    }
+
+    publishing {
+        publications {
+            dockerPublication(MavenPublication) {
+                from components.docker
+                artifactId project.name + "-docker"
+            }
+        }
+    }
+
+The above configuration adds a Maven publication that specifies dependencies
+on `barmodule` and the `someSubProject` Gradle sub project. The resulting POM
+file has two `dependency` entries, one for each dependency.
+
+### Generating docker-compose.yml files from templates
+The `generateDockerCompose` task performs two operations: First, it generates
+a mapping `group:name --> version` from the dependencies of the `docker`
+configuration (see above). Second, it replaces all occurrences of version
+variables of the form `{{group:name}}` in the `docker-compose.yml.template` file
+by the resolved versions and writes the resulting file as `docker-compose.yml`.
+
+**Example**
+
+Assume a `docker-compose.yml.template` as follows:
+
+    myservice:
+      image: 'repository/myservice:latest'
+    otherservice:
+      image: 'repository/otherservice:{{othergroup:otherservice}}'
+
+The `build.gradle` of this project publishes a docker Maven artifact declaring
+a dependency on a docker image published as 'othergroup:otherservice' in version
+0.1.2:
+
+    plugins {
+        id 'maven-publish'
+        id 'com.palantir.docker'
+    }
+
+    docker {
+        name 'foo'
+    }
+
+    group 'mygroup'
+    name 'myservice'
+    version '2.3.4'
+
+    dependencies {
+        docker 'othergroup:otherservice:0.1.2'
+    }
+
+    publishing {
+        publications {
+            dockerPublication(MavenPublication) {
+                from components.docker
+                artifactId project.name + "-docker"
+            }
+        }
+    }
+
+The `generateDockerCompose` task creates a `docker-compose.yml` as follows:
+
+    myservice:
+      image: 'repository/myservice:latest'
+    otherservice:
+      image: 'repository/otherservice:0.1.2'
+
+The `generateDockerCompose` task fails if the template file contains variables that
+cannot get resolved using the provided `docker` dependencies. Version conflicts between
+transitive dependencies of the same artifact are handled with the standard Gradle
+semantics: each artifact is resolved to the highest declared version.
 
 Tasks
 -----
@@ -44,9 +144,14 @@ Tasks
  * `docker`: build a docker container with the specified name and Dockerfile
  * `dockerPush`: push the specified container to a docker repository
  * `dockerPrepare`: prepare to build a docker container by copying
-   dependent task outputs, referenced files, and `dockerfile` into a temporary 
+   dependent task outputs, referenced files, and `dockerfile` into a temporary
    directory
  * `dockerClean`: remove temporary directory associated with the docker build
+ * `dockerfileZip`: builds a ZIP file containing the configured Dockerfile
+ * `generateDockerCompose`: Populates a docker-compose file template with image
+   versions declared by dependencies
+
+
 
 License
 -------
--- a/src/main/groovy/com/palantir/gradle/docker/DockerComponent.java
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerComponent.java
@@ -1,0 +1,44 @@
+package com.palantir.gradle.docker;
+
+import org.gradle.api.artifacts.DependencySet;
+import org.gradle.api.artifacts.ModuleDependency;
+import org.gradle.api.artifacts.PublishArtifact;
+import org.gradle.api.internal.component.SoftwareComponentInternal;
+import org.gradle.api.internal.component.Usage;
+
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+public class DockerComponent implements SoftwareComponentInternal {
+    private final Usage runtimeUsage = new RuntimeUsage();
+    private final LinkedHashSet<PublishArtifact> artifacts = new LinkedHashSet<>();
+    private final DependencySet runtimeDependencies;
+
+    public DockerComponent(PublishArtifact dockerArtifact, DependencySet runtimeDependencies) {
+        artifacts.add(dockerArtifact);
+        this.runtimeDependencies = runtimeDependencies;
+    }
+
+    public String getName() {
+        return "docker";
+    }
+
+    public Set<Usage> getUsages() {
+        return Collections.singleton(runtimeUsage);
+    }
+
+    private class RuntimeUsage implements Usage {
+        public String getName() {
+            return "runtime";
+        }
+
+        public Set<PublishArtifact> getArtifacts() {
+            return artifacts;
+        }
+
+        public Set<ModuleDependency> getDependencies() {
+            return runtimeDependencies.withType(ModuleDependency.class);
+        }
+    }
+}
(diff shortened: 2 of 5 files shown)


uschi2000:
@markelliot

markelliot:
Still digging through this, but one early thought here is if this should all be part of the same plugin?

readme.md
```diff
@@ -33,20 +33,120 @@ Configuration specifying all parameters:
...
+file has two `dependency` entries, one for each dependency.
+
+### Generating docker-compose.yml files from templates
+The `generateDockerCompose` task performs two operations: First, it generates
+a mapping `group:name --> version` from the dependencies of the `docker`
+configuration (see above). Second, it replaces all occurrences of version
+variables of the form `{{group:name}}` in the `docker-compose.yml.template` file
+by the resolved versions and writes the resulting file as `docker-compose.yml`.
+
+** Example **
```
markelliot: `**Example**` (no spaces)

readme.md
```diff
@@ -33,20 +33,120 @@ Configuration specifying all parameters:
...
+    name 'myservice'
+    version '2.3.4'
+
+    dependencies {
+        docker 'othergroup:otherservice:0.1.2'
+    }
+
+    publishing {
+        publications {
+            dockerPublication(MavenPublication) {
```
markelliot, line 122: should we do something like make a method called `addDockerPublication()` that can be referenced in this block vs. folks repeating the publication information?

readme.md
```diff
@@ -33,20 +33,120 @@ Configuration specifying all parameters:
...
+        }
+    }
+
+The `generateDockerCompose` task creates a `docker-compose.yml` as follows:
+
+    myservice:
+      image: 'repository/myservice:latest'
+    otherservice:
+      image: 'repository/otherservice:0.1.2'
+
```
markelliot, line 135: what happens if a `{{group:name}}` block isn't resolved?

what happens in the case of dependency conflicts?

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
```diff
@@ -57,36 +77,34 @@ class DockerExtension {
...
+
+    File getResolvedDockerComposeFile() {
+        return resolvedDockerComposeFile
+    }
+
     public Set<String> getResolvedFiles() {
         return resolvedFiles
     }
 
     public void resolvePathsAndValidate(File projectDir) {
```
markelliot: you don't need `projectDir` anymore I guess?

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -17,34 +17,53 @@ package com.palantir.gradle.docker
...
             dependsOn prepare
         })
 
-        Task push = project.tasks.create('dockerPush', Exec, {
+        Exec push = project.tasks.create('dockerPush', Exec, {
             description = "Pushes Docker image to configured Docker Hub."
             dependsOn exec
         })
 
+        Zip dockerfileZip = project.tasks.create('dockerfileZip', Zip, {
```
markelliot, line 53: this isn't documented in the readme

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -72,6 +91,33 @@ class PalantirDockerPlugin implements Plugin<Project> {
                 workingDir dockerDir
                 commandLine 'docker', 'push', ext.name
             }
+
+            dockerfileZip.with {
+                from(ext.resolvedDockerfile)
+            }
+
+            // Configure docker-compose templating
+            if (ext.resolvedDockerComposeTemplate.exists()) {
```
markelliot, line 101: I think there likely needs to be two states here -- if you configure the template file but it doesn't exist, we likely need to error so that users get reasonable feedback

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -72,6 +91,33 @@ class PalantirDockerPlugin implements Plugin<Project> {
...
+                def dockerDependencies = project.configurations.docker.resolvedConfiguration.resolvedArtifacts
+                def templateTokens = dockerDependencies.collectEntries {
+                    def version = it.moduleVersion.id
+                    [("{{${version.group}:${version.name}}}"): version.version]
+                }
+
```
markelliot, line 107: we need to check here that there are no unresolved deps

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -72,6 +91,33 @@ class PalantirDockerPlugin implements Plugin<Project> {
...
+                generateDockerCompose.with {
+                    from(ext.resolvedDockerComposeTemplate)
+                    into(ext.resolvedDockerComposeFile.parentFile)
```
markelliot, line 110: i think we want to copy into a buildDir?

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -72,6 +91,33 @@ class PalantirDockerPlugin implements Plugin<Project> {
...
+                    rename { fileName -> fileName.replace(
+                        ext.resolvedDockerComposeTemplate.name, ext.resolvedDockerComposeFile.name) }
```
markelliot, line 112: probably want to copy to a standard output name

readme.md
uschi2000: done

readme.md
uschi2000, line 122: the only thing that I feel is redundant/repeated is the artifactId. I wonder if we should in fact split up the docker tag to generate group/module/version by convention: docker tags are `[REGISTRYHOST/][USERNAME/]NAME[:TAG]` and could be split into group=REGISTRYHOST, module=[USERNAME/]NAME, version=TAG. that would mean the docker-compose template would look something like

otherservice:
  image: '{{repository:otherservice}}'

and {{repository:otherservice}}' would get resolved to repository/imagename:version.

readme.md
uschi2000, line 135: - if it doesn't resolve it doesn't get replaced
- dependency conflicts are handled the standard gradle way: upgrade to the lexicographically largest of the transitive versions, or fail if you explicitly set configuration.docker.resolutionStrategy.failOnVersionConflict()

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
uschi2000: done

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
uschi2000, line 53: it is under Tasks.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
uschi2000, line 101: done: will error whenever you set something non-existing explicitly.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
uschi2000, line 107: done

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
uschi2000, line 110: IMO the most likely flow is to have a docker-compose.yml.template in the root project's dir and generate a docker-copmose.yml at the same level, likely .gitignore'd

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
uschi2000, line 112: why not give optionality? default is docker-compose.yml, can override to custom location.

readme.md
markelliot, line 122: So I'm proposing the whole:

```
dockerPublication(MavenPublication) {
    from components.docker
    artifactId project.name + "-docker"
}
```

is likely to be repetitive for anyone that wants to use this, and that we could define `addDockerPublication()` as a function that just does what that block does.

I like the idea of group = registryhost, module = name, version = tag.

readme.md
markelliot, line 135: Should add these to the README

readme.md
```diff
@@ -33,20 +33,120 @@ Configuration specifying all parameters:
...
 -----
 
  * `docker`: build a docker container with the specified name and Dockerfile
  * `dockerPush`: push the specified container to a docker repository
  * `dockerPrepare`: prepare to build a docker container by copying
-   dependent task outputs, referenced files, and `dockerfile` into a temporary 
+   dependent task outputs, referenced files, and `dockerfile` into a temporary
    directory
  * `dockerClean`: remove temporary directory associated with the docker build
+ * `dockerfileZip`: builds a ZIP file containing the configured Dockerfileo
```
markelliot: extra 'o' at the end of the sentence?

src/main/groovy/com/palantir/gradle/docker/DockerComponent.java
```diff
@@ -0,0 +1,44 @@
...
+import org.gradle.api.artifacts.PublishArtifact;
+import org.gradle.api.internal.component.SoftwareComponentInternal;
+import org.gradle.api.internal.component.Usage;
+
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+public class DockerComponent implements SoftwareComponentInternal {
+  private final Usage runtimeUsage = new RuntimeUsage();
```
markelliot: everything else in this project is 4-space.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
markelliot, line 53: sorry, missed it =/

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
markelliot, line 110: It differs from the 'Dockerfile' semantics, but that's okay I guess. The rest of our gradle work tends to produce output results in buildDir, it's a bit strange to break that convention but I understand why you're breaking it.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
markelliot, line 112: pattern matching Dockerfile, but nvm

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -170,6 +173,192 @@ class PalantirDockerPluginTests extends Specification {
...
+        when:
+        BuildResult buildResult = with('publishToMavenLocal').build()
+        then:
+        buildResult.task(':publishToMavenLocal').outcome == TaskOutcome.SUCCESS
+        def publishFolder = new DefaultLocalMavenRepositoryLocator(
+            new DefaultMavenSettingsProvider(new DefaultMavenFileLocations())).localMavenRepository.toPath()
+            .resolve("testgroup")
+
+        // Check java publication has the right dependencies
+        def javaPublisFolder = publishFolder.resolve(temporaryFolder.root.name).resolve("2.3.4")
```
markelliot: javaPublisFolder -> javaPublis**h**Folder

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
uschi2000, line 112: it's a little different from Dockerfile: the former is optional but must exist if the user specifies it, the latter is required in either case.

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
uschi2000: done

readme.md
uschi2000, line 122: Re repetitive: I'd leave the MavenPublication explicit (so that you can use IvyPublication instead if you like) and provide something like:

dockerPublication(MavenPublication) {
  configureDockerDependencyPublication(it)
}

where configureDockerDependencyPublication(it) applies all the defaults. then you can still do the explicit thing if you like.

readme.md
uschi2000, line 135: done

readme.md
uschi2000: done

src/main/groovy/com/palantir/gradle/docker/DockerComponent.java
uschi2000: done

uschi2000:
Responded to all comments, I believe. Will work on providing the configureDockerDependencyPublication() shortcut and the group/module <-> registry/image-name thing in a follow-up commit I suggest.

markelliot:
Seems fine -- you just have a failing test, though.

uschi2000:
yep, just saw that.

uschi2000:
done now.

readme.md
uschi2000, line 122: So it turns out the "group = registryhost, module = name, version = tag" doesn't work since maven artifact ids are restricted to [A-Za-z0-9_-.]+)  :(

PR #15, opened 2015-12-05
Separate PalantirDockerPlugin and DockerComposeTemplatePlugin into tw…

uschi2000:
…o plugins


uschi2000:
@markelliot

uschi2000:
Forgot to update the readme, will do before merging.

src/main/groovy/com/palantir/gradle/docker/DockerComposeTemplatePlugin.groovy
```diff
@@ -0,0 +1,54 @@
...
+import com.google.common.base.Preconditions
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.api.tasks.Copy
+
+class DockerComposeTemplatePlugin implements Plugin<Project> {
+    @Override
+    void apply(Project project) {
+        DockerComposeTemplateExtension ext =
+            project.extensions.create('dockerComposeTemplate', DockerComposeTemplateExtension, project)
```
markelliot: I think this probably ought to be `dockerCompose` instead of `dockerComposeTemplate`

src/main/resources/META-INF/gradle-plugins/com.palantir.docker-compose-template.properties
```diff
@@ -0,0 +1 @@
+implementation-class=com.palantir.gradle.docker.DockerComposeTemplatePlugin
```
markelliot: I think maybe we should call this `com.palantir.docker-compose` (the `-template` is superfluous?)

markelliot:
lgtm, nits on naming of things

uschi2000:
Renamed and updated readme.md . This is good to go from my perspective.

readme.md
```diff
@@ -33,34 +40,22 @@ Configuration specifying all parameters:
...
-used in orchestrating `docker-compose` environments. Each project can declare
-its dependencies on other docker images and publish an artifact advertising those
-dependencies. The plugin uses standard Maven/Ivy machanism for declaring and
-resolving dependencies.
-
-The `generateDockerCompose` task generates a `docker-compose.yml` file from
-a user-defined template by replacing each version variable by the concrete version
-declared by the transitive dependencies of the docker configuration.
-
 ### Specifying dependencies on Docker images
```
markelliot: this really goes with the other plugin docs that start on line 94? (the docker configuration only exists when the `com.palantir.docker-compose` plugin is applied, if I'm reading the other source correctly)

markelliot: er, it looks like you get the configuration either way. I'm not sure it should be created when only applying the `com.palantir.docker` plugin.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -27,6 +26,7 @@ import org.gradle.api.tasks.Exec
 import org.gradle.api.tasks.bundling.Zip
 
 class PalantirDockerPlugin implements Plugin<Project> {
+    @Override
     void apply(Project project) {
         DockerExtension ext = project.extensions.create('docker', DockerExtension, project)
         project.configurations.create("docker")
```
markelliot, line 32: per above, I think we should remove this.

markelliot:
Are there implications for how this project publishes to the Gradle plugins site?

readme.md
uschi2000: you need the 'docker' configuration in both cases: the 'docker' plugin uses it to publish declared dependencies, the 'docker-compose' plugin uses it to resolve declared dependencies. so yes, I think both plugins should add it.

uschi2000:
added block for plugin-publishing of docker-compose plugin.

readme.md
markelliot: I think the dependencies stuff is really just for the docker-compose world, why would you need it in the simple case of just producing a container?

uschi2000: If you want to enable consumers of your container to resolve version strings, then you need to drop some versioned artifact to Ivy/Maven. Even if your project does not use docker-compose.

uschi2000: (Will be in the office this PM, happy to discuss in person.)

markelliot: Ok, cool. So I think this section should really cover the publishing instead of the consuming part, and that this section ought to move to the docker-compose section. Alternatively, we should describe, slightly more abstract than covered here, the general setup for publishing/resolution, and how they're used.

uschi2000: Re-organized the sections.

uschi2000:
Ping.

src/main/groovy/com/palantir/gradle/docker/DockerComposePlugin.groovy
```diff
@@ -0,0 +1,54 @@
...
+            ext.resolvePathsAndValidate()
+            if (ext.resolvedDockerComposeTemplate.exists()) {
+                def dockerDependencies = project.configurations.docker.resolvedConfiguration.resolvedArtifacts
+                def templateTokens = dockerDependencies.collectEntries {
+                    def version = it.moduleVersion.id
+                    [("{{${version.group}:${version.name}}}"): version.version]
+                }
+
+                println(dockerDependencies)
+                println(templateTokens)
```
markelliot: oooh, printlns

src/main/groovy/com/palantir/gradle/docker/DockerComposePlugin.groovy
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /** Replaces all occurrences of templatesTokens's keys by their corresponding values in the given line. */
+    static def replaceAll(String line, Map<String, String> templateTokens, DockerComposeExtension ext) {
+        templateTokens.each { mapping -> line = line.replace(mapping.key, mapping.value) }
+        def unmatchedTokens = line.findAll(/\{\{.*\}\}/)
+        Preconditions.checkState(unmatchedTokens.size() == 0,
+            "Failed to resolve Docker dependencies declared in %s: %s. Known dependencies: %s",
+            ext.resolvedDockerComposeTemplate, unmatchedTokens, templateTokens)
+        line
```
markelliot: would prefer the explicit `return` here (since you're editing the printlns above already)

uschi2000: done

src/main/groovy/com/palantir/gradle/docker/DockerComposePlugin.groovy
uschi2000: ouch. done.

PR #17, opened 2015-12-26
Upgrade to Gradle 2.10

markelliot:


src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
```diff
@@ -65,8 +65,8 @@ class DockerComposePluginTests extends AbstractPluginTest {
         when:
         BuildResult buildResult = with('generateDockerCompose', '--stacktrace').buildAndFail()
         then:
-        buildResult.standardError.contains("Failed to resolve Docker dependencies declared in")
-        buildResult.standardError.contains("{{foo:bar}}")
+        buildResult.output.contains("Failed to resolve Docker dependencies declared in")
```
uschi2000, line 68: I understand that the plugin will still work with <2.10, but the stderr/stdout semantics changed for 2.10, hence this change?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.8-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.10-bin.zip
```
uschi2000, line 5: with the 2.10 upgrade, can you get rid of some of that testClasses business in the tests?

src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
markelliot, line 68: right

gradle/wrapper/gradle-wrapper.properties
markelliot, line 5: there's not a meaningful amount of change, still: https://docs.gradle.org/current/userguide/test_kit.html

uschi2000:
Lgtm

PR #18, opened 2015-12-26
Switch to Circle CI

markelliot:


src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -157,7 +157,7 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
         buildResult.task(':dockerPrepare').outcome == TaskOutcome.SUCCESS
         buildResult.task(':docker').outcome == TaskOutcome.SUCCESS
         exec("docker inspect --format '{{.Author}}' ${id}") == "'${id}'\n"
-        exec("docker rmi ${id}")
+        exec("docker rmi -f ${id}")
```
markelliot, line 160: saw a consistent failure without the force flag when building this on circle (though strangely, only on this particular test); the `rmi` is just for cleanup, so happy to just punch the force button.

PR #19, opened 2016-01-17
Add support for creating and pushing tags

markelliot:


markelliot:
@jmcampanini: I think a lot of our gradle configs are including dockerPushLatest, this is a generalized way to deal with those setups, and you might think about converting highline to match

jmcampanini:
thank you, sir! :) when we look through the highline docker/ete stuff, we'll look to using the standard libraries we built.

PR #24, opened 2016-01-25
Add docker-compose start/stop tasks

michael-yx-wu:


uschi2000:
So the only benefit of this thing is the ability to depend on the docker build task?

michael-yx-wu:
You can optionally use the `generateDockerCompose` task. You can also directly just call `dockerComposeUp` if you don't need it. It doesn't depend on a docker image being built by the `docker` plugin in this repo, if that's what you meant?

markelliot:
I think there's not quite enough configuration here for dockerComposeUp to be valuable over just calling `docker-compose up`.

src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
```diff
@@ -130,4 +130,32 @@ class DockerComposePluginTests extends AbstractPluginTest {
...
+
+        when:
+        BuildResult buildResult = with('dockerComposeUp', 'dockerComposeStop').build()
+        BuildResult offline = with('dockerComposeRemove').build()
+
+        then:
+        buildResult.task(':dockerComposeUp').outcome == TaskOutcome.SUCCESS
+        buildResult.task(':dockerComposeStop').outcome == TaskOutcome.SUCCESS
+
+        offline.task(':dockerComposeRemove').outcome == TaskOutcome.SUCCESS
```
markelliot, line 175: I think we need to find a way to test that the containers actually came up, then stopped, then removed in this test

michael-yx-wu:
Ashwin suggested adding the docker compose tasks in compass to the plugin to avoid duplication of such tasks across projects. Should I make it easier for users to add additional flags to docker-compose up?

markelliot:
Interesting. My concern is that we have one product now that looks like this for setup:

```
docker-compose -f docker-compose.yml --x-networking -p testnetwork up –d
```

And there's not really a way to provide the networking option. It seems like a bit of a pit of features to keep pace with how docker-compose sends arguments (the docker build commands are incredibly stable, which is why I opted to shell out in the other plugins here).

uschi2000:
Sharing Mark's concerns; task dependency only doesn't justify the feature skin-hole that this will become.

michael-yx-wu:
We could add a configuration block that allows users to specify additional arguments as well as whether the arguments apply to docker-compose or to a particular docker-compose command. It's unlikely people will want to add arguments to stop/remove. So we could do something like:

```
dockerCompose {
    composeFlags '--networking -p my-project'
    upFlags '--force-recreate'
}
```

markelliot:
I'm just worried about what it's going to take to capture the full set of arguments in a reasonable way — docker-compose is still under pretty heavy churn in terms of CLI API, I believe.

michael-yx-wu:
Okay, that makes sense. I'm closing this pull request for now then. We can revisit when we feel that the cli docker-compose is less volatile.

PR #28, opened 2016-02-09
Remove force flag from docker tag

michael-yx-wu:
`-f` is deprecated in Docker 1.10.0 and will be removed in 1.12.0. I'm happy to sit on this PR if it's too early.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
@@ -103,7 +103,7 @@ class PalantirDockerPlugin implements Plugin<Project> {
                         group = 'Docker'
                         description = "Tags Docker image with tag '${tagName}'"
                         workingDir dockerDir
-                        commandLine 'docker', 'tag', '--force=true', ext.name, computeName(ext.name, tagName)
+                        commandLine 'docker', 'tag', ext.name, computeName(ext.name, tagName)
                         dependsOn exec
                     })
                     tag.dependsOn subTask
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,7 @@ machine:
   java:
     version: openjdk7
   pre:
-    - curl -sSL https://s3.amazonaws.com/circle-downloads/install-circleci-docker.sh | bash -s -- 1.10.3
+    - curl -sSL https://s3.amazonaws.com/circle-downloads/install-circleci-docker.sh | bash -s -- 1.10.0
     - sudo wget -O /usr/local/bin/docker-compose 'https://github.com/docker/compose/releases/download/1.6.2/docker-compose-Linux-x86_64'
     - sudo chmod 0755 /usr/local/bin/docker-compose
   services:
@@ -24,4 +24,3 @@ deployment:
     commands:
       - ./gradlew -i bintrayUpload
       - ./gradlew -i -Dgradle.publish.key=$GRADLE_KEY -Dgradle.publish.secret=$GRADLE_SECRET publishPlugins
-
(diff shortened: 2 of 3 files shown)


michael-yx-wu:
It looks like the build failed CircleCI is running an older version of Docker. I can wait until we're ready to move to 1.10.0.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -103,7 +103,7 @@ class PalantirDockerPlugin implements Plugin<Project> {
                         group = 'Docker'
                         description = "Tags Docker image with tag '${tagName}'"
                         workingDir dockerDir
-                        commandLine 'docker', 'tag', '--force=true', ext.name, computeName(ext.name, tagName)
+                        commandLine 'docker', 'tag', ext.name, computeName(ext.name, tagName)
```
uschi2000, line 106: wouldn't this fail when you try tagging things multiple times (e.g. ":latest")?

markelliot, line 106: Right, we need to keep `--force=true` to deal with the existence of containers locally from prior builds. I'd prefer to keep as-is.

michael-yx-wu, line 106: I don't think you'll have that issue in 1.10: https://docs.docker.com/engine/deprecated/#f-flag-on-docker-tag:9c0d3b2efa16c5c274427efa73f212c9. But since most people are probably on 1.9, we shouldn't merge it just yet

markelliot, line 106: Ok, can keep this PR open now, but don't want to merge.

dataethos, line 106: Docker Engine users pulling base images from Docker Hub ought to be on =>1.10 now, to avoid image ID incompatibilities.  Would you re-consider merging this PR?

I'm building my images with Atlassian Bamboo.  `gradle dockerTag` returns with a 0 exit code, but the warning message from `docker` outputs to stderr.  It seems that Bamboo is picking up the output on stderr and fails the build, even though the `gradle` command succeeds.

I created an `Exec` task in my build based on the code in `PalantirDockerPlugin` without the `--force=true` option, and my Bamboo build now succeeds.

**Environment**
- **OS:** CentOS 7
- **Docker Engine:** 1.10.2
- **Gradle**: 2.10 wrapper
- **Bamboo:** 5.9.9

markelliot, line 106: Yeah, I think we should probably merge this now...but it needs to be brought up to date/checks need to pass.

michael-yx-wu, line 106: Cool. I'll rebase!

michael-yx-wu:
Build passes locally (I'm on docker 1.10.3 and compose 1.6.2), but seems to fail on Circle with the following error:

```
com.palantir.gradle.docker.PalantirDockerPluginTests > check files are correctly added to docker context FAILED
    Condition not satisfied:

    |                        |
    false                    a661be69-3ab8-4182-ad7a-33fae331d9e4
        at com.palantir.gradle.docker.PalantirDockerPluginTests.check files are correctly added to docker context(PalantirDockerPluginTests.groovy:160)
Gradle Test Executor 2 finished executing tests.
```

michael-yx-wu:
The `execCond(...) || true` change is a temporary workaround to #35

PR #31, opened 2016-04-07
Upgrade docker to 1.10 and compose to 1.6.2

michael-yx-wu:


markelliot:
Upgraded to 1.10.3 instead of 1.10.0 in #32

Issue #36, opened 2016-04-18
remove or make DockerSetupTask skippable


dzwicker:
With the new Docker for Mac OS X and Windows the environment variables set by `docker-machine` aren't used any more.

jmcampanini:
+1 to skippable. i'm just adding `-x dockerSetup` and it works for me

antoineco:
I guess #56 solved that.

PR #37, opened 2016-04-18
New docker Mac OS X and Windows don't relay on docker-machine

dzwicker:
New docker Mac OS X and Windows versions don't depend on docker-machine any more. So we can remove the DockerSetupTask. See #36


build.gradle
```diff
@@ -41,23 +41,37 @@ if (System.env.CIRCLE_TEST_REPORTS) {
...
             from components.java
             artifact(sourceJar)
             artifact(publishPluginJavaDocsJar)
         }
     }
+    repositories {
+        maven {
+            credentials {
+                username 'deployment'
+                password 'MXJmqxvTGve9s6q9y7Ks'
```
markelliot, line 68: pretty sure you don't want to commit the set of changes to this file, and you might want to think about changing this exposed password.

markelliot:
Given that the Docker for Mac isn't widely available yet, I'm not sure it's safe to make this change. I'd be open to having an extra configuration item in the extension to enable/disable the setup task, though.

jkozlowski:
I hit the same problem, as a workaround running with -x dockerSetup works fine for me. In general I think it'd be better if we ran something like docker ps and checked return code to see if we can connect to docker: that would potentially work with docker-machine and docker for mac beta.

markelliot:
I think test running docker ps makes sense, happy to see that in a PR.

SerialVelocity:
[dlite](https://github.com/nlf/dlite) also fails to work with this too.

uschi2000:
Docker for Mac is now a public beta, so let's make some progress here. @jkozlowski will you drive this or should I?

jkozlowski:
see #55.

markelliot:
Took #56 as a clean version of this.

PR #38, opened 2016-04-19
Add more options to Docker Run command

mglazer:
- Support volumes
- Support running containers which don't need to be daemonized

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/DockerRunPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerRunPlugin.groovy
@@ -19,6 +19,12 @@ import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.tasks.Exec
+import org.gradle.logging.StyledTextOutput
+import org.gradle.logging.StyledTextOutputFactory
+import static org.gradle.logging.StyledTextOutput.Style
+
+import java.io.File
+import java.util.Map.Entry
 
 import com.google.common.collect.Lists
 
@@ -40,7 +46,6 @@ class DockerRunPlugin implements Plugin<Project> {
             group = 'Docker Run'
             description = 'Runs the specified container with port mappings'
             dependsOn setup
-            finalizedBy dockerRunStatus
         })
 
         Exec dockerStop = project.tasks.create('dockerStop', Exec, {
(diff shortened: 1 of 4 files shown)


mglazer:
```
Error removing intermediate container 5376e7877f64: rmdriverfs: Driver btrfs failed to remove root filesystem 5376e7877f64770395556c3d5fd79d2f82277e2f25d2c56dc742625b748fdd64: Failed to destroy btrfs snapshot /var/lib/docker/btrfs/subvolumes for 3084ef7d1cb27bb729dd59d082cf6fb5ccf4022a72d40eddb7f5789003325a95: operation not permitted
```

So I can commit this without tests, because this seems like a CircleCI issue with an inability to clean up containers with attached volumes:

https://discuss.circleci.com/t/docker-error-removing-intermediate-container/70/10

The tests run fine locally.

mglazer:
@markelliot : given that integration testing with mounted volumes fails on CircleCI...do you propose I just commit without the tests? Or write less complete unit tests instead?

markelliot:
Any ideas why it fails on Circle? We can install a non-Circle Docker in the setup if that'd fix it.

If no ideas, let's keep the tests but ignore them for Circle only and record an issue.

Sent from my iPhone

On Apr 22, 2016, at 8:28 AM, Mike Glazer <notifications@github.com<mailto:notifications@github.com>> wrote:

@markelliothttps://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_markelliot&d=CwMCaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=Vp81aOxWZvuVgVK_wp-VF3pIYG92B19LcCw6XKeYC0U&m=BbfoUW1NxTYCHnzsvjwfFew-19EKTOHzKJVtY2x2unk&s=r6drCDPWDe-DEScngSe9U5vgd2A-vwuHHHDOdeUMVN4&e= : given that integration testing with mounted volumes fails on CircleCI...do you propose I just commit without the tests? Or write less complete unit tests instead?

mglazer:
See the linked docker error removing container:

https://discuss.circleci.com/t/docker-error-removing-intermediate-container/70/10

Basically, it seems like you can't test with mounted volumes. There's another issue with the integration test because the temp folder is created outside of the /Users, which on Windows/OSX will fail because of how Docker works in those environments. I already have a check for that though.

I've tested this in an actual setup, so I know it works (ie: published locally and tested). 

I'll add an IF_CIRCLE test.

mglazer:
For future commits, we may want clean: true to be the default, since it seems the containers aren't cleaning up after themselves, hence why I created one as `bar-nodaemonize`

src/main/groovy/com/palantir/gradle/docker/DockerRunPlugin.groovy
```diff
@@ -73,11 +78,30 @@ class DockerRunPlugin implements Plugin<Project> {
...
+                  finalizedBy dockerRunStatus
+                }
                 for (String port : ext.ports) {
                     args.add('-p')
                     args.add(port)
                 }
+                for (Entry<String,String> volume : ext.volumes.entrySet()) {
+                    File localFile = new File(project.projectDir, volume.key)
+
+                    if (!localFile.exists()) {
```
markelliot, line 97: It might make sense to throw on this to prevent Docker from making the folder in a VM (which is what happens on Macs and becomes quite annoying to deal with )

markelliot:
Seems fine, one thought re: error states and then squash?

src/main/groovy/com/palantir/gradle/docker/DockerRunPlugin.groovy
mglazer, line 97: Ya, I went back and forth on that one, since it wasn't strictly an error like the port check you had in the extension. If you think it's reasonable to move there, I will.

mglazer:
Added louder error (currently, no way to work around it, so this is a fail hard error), and squashed all commits together.

Issue #45, opened 2016-05-04
Feature Request: Support for --build-arg in docker task


rmelick:
I'm attempting to build a docker image that contains a versioned tar (kiwiCloud-6.0.0.0-SNAPSHOT.tar) file that is produced by a different part of my build.

build.gradle file

```
plugins {
    id 'com.palantir.docker' version '0.7.0'
}

docker {
    name "docker.com/rmelick/cloud/api:$version"
    dependsOn tasks.getByPath(':distTar')
}

```

Dockerfile

```
FROM java:8

ADD kiwiCloud-* /
```

I would like to be able to pass the version parameter into my Dockerfile as a build-arg.  This would let me remove the wildcard from the ADD command, and would also let me use the version when referencing the contents of the tar (the script I want to use for ENTRYPOINT is inside the tar).

markelliot:
A couple of ways you could do this:
- use Docker arguments (https://docs.docker.com/engine/reference/builder/#arg)
- use Gradle substitution to generate your Dockerfile (and configure this plugin to point at the generated file)

I'm not sure it makes sense for this plugin to become involved in modifying your Dockerfile, though.

rmelick:
@markelliot I agree, I think using the docker --build-arg argument would be the right way to go, rather than modifying the Dockerfile.  I'll try to send a PR today.

markelliot:
Released in 0.7.1

PR #46, opened 2016-05-11
#45: Add support for docker build args

rmelick:
I added a new buildArgs parameter to the docker configuration and a new test.  buildArgs lets you control the --build-arg parameter to docker build, as requested in #45.

I am more than happy to add more tests, squash the branch down into a single commit, make the commit messages fit a certain style, etc.

I also fixed a test for Docker Run that has likely been failing for a while, but is only enabled on local Linux environments.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -35,6 +35,8 @@ automatically include outputs of task dependencies in the Docker build context.
 - `dependsOn` (optional) an argument list of tasks that docker builds must depend on;
   defaults to the empty set
 - `files` (optional) an argument list of files to be included in the docker build context
+- 'buildArgs' (optional) an argument map of string to string which will set --build-arg;
+  arguments to the docker build command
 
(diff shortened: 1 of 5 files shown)


markelliot:
Thanks, this looks great. Could I get you to sign Palantir's CLA so we can accept the contribution?

Fill out the an appropriate Contributor License Agreement and send it to opensource@palantir.com:
- [Individual Contributor License Agreement](https://github.com/palantir/auth-tokens/blob/develop/palantir_individual_cla.pdf?raw=true)
- [Corporate Contributor License Agreement](https://github.com/palantir/auth-tokens/blob/develop/palantir_corporate_cla.pdf?raw=true)

rmelick:
Thanks @markelliot .  Just sent it in.

markelliot:
Thanks!

PR #47, opened 2016-05-11
Fix a couple of breaking unit tests

antony:
I had a couple of consistently failing unit tests, which this PR fixes.

My environment is:

Ubuntu 15.10
Docker version 1.9.1, build a34a1d5

The fixes I made were:
- Test starts docker image 'foo' but expects 'bar' to stop.
- Tagging test hits a conflict when trying to tag 'latest'. I think 'latest' is a default tag, so I've made the test less likely to conflict by changing the built tag names.

Please let me know if these fixes are valid as I intend to develop a new feature and I'll be using my fork as a base.

### DIFF
--- a/src/test/groovy/com/palantir/gradle/docker/DockerRunPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/docker/DockerRunPluginTests.groovy
@@ -16,7 +16,6 @@
 package com.palantir.gradle.docker
 
 import org.gradle.testkit.runner.BuildResult
-import org.gradle.testkit.runner.TaskOutcome
 
 import static org.gradle.testkit.runner.TaskOutcome.SUCCESS
 
--- a/src/test/groovy/com/palantir/gradle/docker/DockerRunPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/docker/DockerRunPluginTests.groovy
@@ -16,8 +16,7 @@
 package com.palantir.gradle.docker
 
 import org.gradle.testkit.runner.BuildResult
-
-import static org.gradle.testkit.runner.TaskOutcome.SUCCESS
+import org.gradle.testkit.runner.TaskOutcome
 
 class DockerRunPluginTests extends AbstractPluginTest {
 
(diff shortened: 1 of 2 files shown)


antony:
I've tried this code on three docker versions now
- The tests pass on Docker 1.10.x
- The first commit makes them additionally pass on Docker 1.9.x
- The last commits make them additionally pass on Docker 1.11.x

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -312,20 +316,22 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
 
             docker {
                 name '${id}'
-                tags 'latest', 'another'
+                tags 'mytag', 'another'
```
markelliot: this really needs to pass as `latest`

src/test/groovy/com/palantir/gradle/docker/DockerRunPluginTests.groovy
```diff
@@ -174,12 +175,12 @@ class DockerRunPluginTests extends AbstractPluginTest {
...
-        buildResult.task(':dockerRemoveContainer').outcome == TaskOutcome.SUCCESS
+        buildResult.task(':dockerRemoveContainer').outcome == SUCCESS
 
-        buildResult.task(':dockerRun').outcome == TaskOutcome.SUCCESS
+        buildResult.task(':dockerRun').outcome == SUCCESS
         buildResult.output =~ /(?m)HELLO WORLD/
-        buildResult.task(':dockerRunStatus').outcome == TaskOutcome.SUCCESS
-        buildResult.output =~ /(?m):dockerRunStatus\nDocker container 'bar' is STOPPED./
+        buildResult.task(':dockerRunStatus').outcome == SUCCESS
+        buildResult.output =~ /(?m):dockerRunStatus\nDocker container 'foo' is STOPPED./
```
markelliot, line 182: as far as I can tell, this is the substantive change in this file, I'd prefer to keep the style change out of something that's looking to fix bugs

antony, line 182: @markelliot Apologies for that - I tend to refactor test classes as I work to understand them. Happy to revert these style changes if need be.

btw - the actual change in the file above is to match the started container name `foo` rather than `bar`

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
antony: I think the latest tag is generated by default in docker < 1.10 - hence you get a conflict, so other than put logic in the test, I'm stumped as to what the solution would be to make this work pre 1.10

src/test/groovy/com/palantir/gradle/docker/DockerRunPluginTests.groovy
markelliot, line 182: No worries, thanks for iterating – keeping style changes/bug fixes separate have generally helped with traceability in the past, so would prefer you save the style changes for another PR if you don’t mind.

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
markelliot: I had this passing on Docker < 1.10 (the CI system ran 1.9 until very very recently), so not sure why this fails. Any chance you could walk through the error you’re encountering?

antony: Sure - will do when I get home.

antony: This appears to work for me now so I've reverted it back to 'latest'. You will note that the test was always testing for the presence of 'latest', so the intent didn't change :)

Perhaps my machine had a stale state from a previous test run - though having the 'cleanup' block should prevent that.

antony:
So after my cleanup, this commit doesn't actually do anything any more - I can only assume the fixes were made on master since my last PR. Closing it!

PR #49, opened 2016-05-18
Publish to Bintray without user intervention

markelliot:


uschi2000:
very nice. is that a new feature?

markelliot:
Saw it on the AtlasDb repo, not sure if it was added recently or I just missed it when I originally set up the publishing blocks.

PR #55, opened 2016-06-20
Run 'docker version' in dockerSetup.

jkozlowski:
- Should print something like the following if docker is down:

```
:dockerSetup
    Client:
    Cannot connect to the Docker daemon. Is the docker daemon running on this host?
     Version:      1.12.0-rc2
     API version:  1.24
     Go version:   go1.6.2
     Git commit:   906eacd
     Built:        Fri Jun 17 20:35:33 2016
     OS/Arch:      darwin/amd64
     Experimental: true
    :dockerSetup FAILED
```

Should fix #37. Looking for ideas on how to test this; I tried sudo service stop docker, but I'm not sure it is actually possible to stop the docker daemon on CircleCI?

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/DockerSetupTask.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerSetupTask.groovy
@@ -16,10 +16,8 @@
 package com.palantir.gradle.docker
 
 import org.gradle.api.DefaultTask
-import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.tasks.TaskAction
-import org.gradle.internal.os.OperatingSystem
 
 class DockerSetupTask extends DefaultTask {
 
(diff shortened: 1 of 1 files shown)


jkozlowski:
Or just remove it altogether, the output is fairly informative from just running the docker commands.

src/main/groovy/com/palantir/gradle/docker/DockerSetupTask.groovy
```diff
@@ -30,32 +28,14 @@ class DockerSetupTask extends DefaultTask {
...
-            if (!System.env.DOCKER_CERT_PATH) {
-                error += 'DOCKER_CERT_PATH not set.\n'
-            }
-            if (!System.env.DOCKER_MACHINE_NAME) {
-                error += 'DOCKER_MACHINE_NAME not set.\n'
-            }
-            if (error != '') {
-                throw new GradleException(error += 'Please make sure your Docker VM is running.')
-            }
+        project.exec {
```
uschi2000, line 36: does this fail with a reasonable error?

jkozlowski, line 36: You get what I pasted in the description and some version information: 

```
Cannot connect to the Docker daemon. Is the docker daemon running on this host?
```

Can change to docker ps, the output will be less noisy. Or in general get rid of this task, all the other ones will fail with the same message when cannot connect.

markelliot, line 36: I think just kill the task entirely.

jkozlowski, line 36: here you go: #56

markelliot:
Took #56 as preferred alternative.

PR #59, opened 2016-07-27
Add 'pull' parameter to docker plugin

antoineco:
This adds a new `pull` parameter to the `docker` plugin, which defines whether Docker should attempt to pull a newer version of the base image before starting a build.

Oh, and I also added syntax highlighting to the Readme, because it looks great on GitHub :relieved:

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
@@ -89,6 +89,9 @@ class PalantirDockerPlugin implements Plugin<Project> {
                     buildCommandLine.addAll('--build-arg', "${buildArg.getKey()}=${buildArg.getValue()}")
                 }
             }
+            if (ext.pull) {
+                buildCommandLine.add '--pull'
+            }
             buildCommandLine.addAll(['-t', ext.name, '.'])
             exec.with {
                 workingDir dockerDir
--- a/src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
@@ -330,7 +330,7 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
         execCond("docker rmi -f ${id}:another")
     }
 
-    def 'build args'() {
+    def 'build args are correctly processed'() {
         given:
         String id = 'id7'
         temporaryFolder.newFile('Dockerfile') << '''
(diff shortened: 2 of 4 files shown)


markelliot:
Thanks, this looks great. Could I get you to sign Palantir's CLA so we can accept the contribution?

Fill out the an appropriate Contributor License Agreement and send it to opensource@palantir.com:
- [Individual Contributor License Agreement](https://github.com/palantir/auth-tokens/blob/develop/palantir_individual_cla.pdf?raw=true)
- [Corporate Contributor License Agreement](https://github.com/palantir/auth-tokens/blob/develop/palantir_corporate_cla.pdf?raw=true)

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -361,4 +361,30 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
...
+                pull true
+            }
+        """.stripIndent()
+
+        when:
+        BuildResult buildResult = with('docker').build()
+
+        then:
+        buildResult.task(':docker').outcome == TaskOutcome.SUCCESS
+        buildResult.output.contains 'Pulling from library/alpine'
```
uschi2000, line 387: this assertion looks bogus: it's true with `pull true` as well as `pull false` if `alpine:3.2` does not exist (which should be the case on a vanilla CI node.

uschi2000:
👎  (see comment on test)

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
antoineco, line 387: I'm aware of that case, but the corresponding image will exist from previous tests. Should I maybe ensure that with an image dependency like

``` gradle
dependencies {
    docker 'alpine:3.2'
}
```

?

uschi2000, line 387: You cannot rely on test ordering. The correct way is to first pull the base
image, then run the `docker` task and assert that it pulls again.

antoineco:
I just added a dependency on the base image in order to ensure it is present on the system when the image is built.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -89,6 +89,9 @@ class PalantirDockerPlugin implements Plugin<Project> {
                     buildCommandLine.addAll('--build-arg', "${buildArg.getKey()}=${buildArg.getValue()}")
                 }
             }
+            if (ext.pull) {
```
uschi2000, line 92: if wonder if we should instead allow for a more general mechanism that can also specify other cmd line flags, etc `--quiet`, etc?

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -361,4 +361,34 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
...
+        String id = 'id8'
+        temporaryFolder.newFile('Dockerfile') << '''
+            FROM alpine:3.2
+        '''.stripIndent()
+        buildFile << """
+            plugins {
+                id 'com.palantir.docker'
+            }
+
+            dependencies {
```
uschi2000: this block is irrelevant for this test; in particular, it doesn't pull the base image before building. please just pull the image explicitly before the test. (see DockerRunPlugin.groovy for how to do this with an Exec task)

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
antoineco, line 92: That would be ideal indeed, but if we follow this logic arguments like `dockerfile` or `buildArgs` can be removed since it's possible to pass them as a map argument. My understanding of Gradle plugins is that they replace scripts and complex command lines by providing user friendly interfaces for common actions. I believe 'pull' is a very common parameter used on CI systems.

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
antoineco: My bad, I thought `docker` dependencies would always be fetched before `docker*` tasks. I completely misunderstood the purpose of that option. Let me amend my commit one more time.

uschi2000:
👍

uschi2000:
Thanks, Antoine!

Issue #58, opened 2016-07-21
using parameter 'files' does not lead to expected results with directories


stefan-huettemann:
Hi & thank you for this nice plugin.

We have an issue (questions) with configuring `docker` using the `files` parameter:

``` javascript
docker {
    name 'myContainer'
    files 'x'
}
```

where `x` is a directory. 

We would expect that the directory `x` (including all its contents) will get copied to the `build/docker` area.

Instead **only the content** of directory `x` gets copied (not the directory `x` itself).

Using the configuration below leads to the result we would expect from the `files` parameter:

``` javascript
dockerPrepare {
    from('.') {
        include 'x'
    }
}
```

_Questions_:
1. Is this a 'feature' of the `files` parameter?
2. Is the configuration as stated above intended  to be used a such by the design of the plugin?

Thank & Regards
-Stefan

markelliot:
We haven't traditionally used this to add directories, so the behavior isn't intentionally one way or the other.

antoineco:
This behavior is due to the way [`CopySpec.from()`](https://docs.gradle.org/current/javadoc/org/gradle/api/file/CopySpec.html#from%28java.lang.Object[]%29) works, which is also reminded in [Chapter 18. Working With Files](https://docs.gradle.org/current/userguide/working_with_files.html#sec:copying_files):

> When an argument resolves to a directory, everything under that directory (but not the directory itself) is recursively copied into the destination directory.

The default behavior of Docker is to send everything that exists in the same directory as the Dockerfile to the build context (excluding what is defined in the `.dockerignore` file). Sending the content of a directory instead of the directory itself is misleading and breaks the 1:1 compatibility with the CLI, since paths differ in the end.

Thanks for posting this workaround @stefan-huettemann, that definitely helped since I face facing a similar issue.

markelliot:
Thanks -- we'd be happy to take a contribution which decides if an entry is a directory and if so copies the whole directory instead of just the file.

stefan-huettemann:
Hi. Please consider a solution as proposed in https://github.com/palantir/gradle-docker/pull/62.

PR #62, opened 2016-08-02
possible fix for https://github.com/palantir/gradle-docker/issues/58

stefan-huettemann:
Hi.

Please consider this possible (quick) solution.

Note: the solution tries to respect the patterns found in .dockerignore but does so in a very simple manner. An improvemnt would be to actually translate the golang filepath patterns used in .dockerignore to gradles "ANT like" exclude pattern.

Regards,
-Stefan

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -36,7 +36,7 @@ automatically include outputs of task dependencies in the Docker build context.
   `${projectDir}/Dockerfile`
 - `dependsOn` (optional) an argument list of tasks that docker builds must depend on;
   defaults to the empty set
-- `files` (optional) an argument list of files to be included in the docker build context
+- `files` (optional) an argument list of files to be included in the docker build context; if this parameter is omitted, all files in `${projectDir}` are included
 - `buildArgs` (optional) an argument map of string to string which will set --build-arg
   arguments to the docker build command; defaults to empty, which results in no --build-arg parameters
 - `pull` (optional) a boolean argument which defines whether Docker should attempt to pull
--- a/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
@@ -79,7 +79,23 @@ class PalantirDockerPlugin implements Plugin<Project> {
                     }
                 }
                 from ext.dependencies*.outputs
-                from ext.resolvedFiles
+                if (!ext.resolvedFiles.isEmpty()) {
+                    // provide compatibility with optional 'files' parameter:
+                    from ext.resolvedFiles
+                } else {
+                    // default: copy all files that are NOT excluded by '.dockerignore'
+                    // NOTE:
+                    //  the patterns from .dockeringore DO NOT EXACTLY match gradles exclude patterns
+                    //  ... but for many cases it provides a reasonable solution :)
+                    // TODO: translate golangs filepath.Match pattern to gradles exclude pattern
+                    File dockerignore = new File("${project.projectDir}/.dockerignore")
+                    from(project.projectDir) {
+                        exclude "${project.buildDir.name}"
+                        if (dockerignore.exists()) {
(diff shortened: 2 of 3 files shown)


uschi2000:
I don't quite understand the `.dockerignore` issue: why wouldn't `docker build` pick up the `.dockerignore` file and interpret it correctly?

uschi2000:
independently, this change needs tests.

stefan-huettemann:
Concering ".dockerignore":
the plugin copies all files to the "build/docker" directory and runs the `docker build` from this directory. The files needed by `docker build` have to be in the build directory for `docker build` to work correctly. The plugin currently solves this by listing all required files in the `ext.files` parameter. This leads to the current issue with not coping the contents of directories.

My suggestion: copy everything found in "${project.projectDir}" to "${project.buildDir}/docker" directory by default (excluding ""${project.buildDir}").

```
if (!ext.resolvedFiles.isEmpty()) {
      // provide compatibility with optional 'files' parameter:
      from ext.resolvedFiles
} else {
      // default: copy all files excluding the project buildDir
      from(project.projectDir) {
        exclude "${project.buildDir.name}"
      }
}
```

This would have the drawback that files and directories would get copied to "build/docker" that actually are excluded in `.dockerignore` and may lead to copying large amout of trash around.

Therefore I suggest to try to respect the `.dockerignore` in the perparation step as good as possible.

If you like that change, let me know and I will provide some tests with it.

uschi2000:
I see, so you're worried about copying too much data into the temp directory. IMO, correctness is more important here than optimizing on the amount of data copied. So: if you can guarantee that the files excluded by your `.dockerfile` interpretation is always a subset of the files excluded by Docker's interpretation, great. Otherwise, let's skip this optimization to avoid bad surprises.

stefan-huettemann:
That seems reasonable. Since I cannot _guarantee_ that gradles exclude is a subset of Dockers ignore, I reduced the change to a simple default to copy everything found in the projectDir excluding the temporary build directory.

Hope that helps :)

markelliot:
Why not just copy the .dockerignore file by default to the temp staging directory? (Then Docker will respect that)

uschi2000:
Yep, that was my suggestion, Mark.

uschi2000:
OK, the functionality looks fine now. This needs some tests though.

stefan-huettemann:
@markelliot: the `.dockerignore` file will get copied to the temporary `build/docker` directory. Therefore the `docker build` works as expected (respecting the `.dockerignore`). The only issue is, that there might be some files/directories getting copied to `build/docker` that eventually get ignored by the docker build. So I think the implementation is correct.

I will check for how to provide tests for this now ...

antony:
This looks like it will work, obviously the tests will prove it :)

You can just use `if (ext.resolvedFiles)` for clarity. In groovy truth, it's the equivalent of `!ext.resolvedFiles.isEmpty()`. It also means you can switch the order of your `if` statement, since checking for a negation is considered undesirable.

stefan-huettemann:
Hi. I provided a test similar to the test for the original `ext.files` parameter and improved the code style.

Concerning the tests I have a side remark: at my current project they gave us nice vintage Windows 7 PCs. (Yes, 7). About half of the existing tests do not run on Windows 7. Everything runs fine on my Mac (using docker toolbox as well). My guess is, that no one ever tried to run these tests on Windows, right? Or am I missing some glue thingy for Windows?

uschi2000:
Cool, that looks fine. Last nit: the readme needs to mention what happens when `files` is not specified.

uschi2000:
(sorry this had slipped off my radar last week, Stefan)

stefan-huettemann:
readme.md is updated.

uschi2000:
Let me go and see what the deal is with the CLA check; have you submitted one of those, Stefan?

stefan-huettemann:
Nope. I haven't submitted any CLA ... but I am just on the way now ..

stefan-huettemann:
... should be done now. I guess.

PR #70, opened 2016-09-27
Create the correct name for images

guenhter:
Just replacing the first colon til the end of the image name is not the correct way to replace a tag.
The name itself has to be checked further and than the correct paths can be split.

Just see the new test-method "check if compute name replaces the name correctly" what the possible values are.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.12-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-bin.zip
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @guenhter! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

markelliot:
is it the case that versions with slashes or colons are disallowed, then?

guenhter:
Yes, if you try for instance
`docker tag hello-world host:5000/hello-world` 
than you get a tag with name `hello-world host:5000/hello-world` and version `latest`

guenhter:
@palantirtech 
I signed the CLA

markelliot:
My question is more around if these are valid:

```
hostname:5000/v1:v/1
hostname:5000/v1:v:1
```

And how docker would treat these with respect to tags.

Maybe worth a link to Docker's spec or actual code.

guenhter:
`docker tag hello-world  hostname:5000/v1:v/1
Error parsing reference: "hostname:5000/v1:v/1" is not a valid repository/tag`

markelliot (approved):

PR #72, opened 2016-09-30
Set --quiet option on builds to suppress some output

markelliot:


src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -91,7 +91,7 @@ class PalantirDockerPlugin implements Plugin<Project> {
                 into dockerDir
             }
 
-            List<String> buildCommandLine = ['docker', 'build']
+            List<String> buildCommandLine = ['docker', 'build', '--quiet']
```
markelliot: @uschi2000 curious for your thoughts here, but builds for some of our projects are generating huge reports -- think we can eliminate a lot of junk build content by setting --quiet here.

uschi2000: Can we inspect whether Gradle was called with --info or --debug and set this accordingly?

markelliot:
That seems pretty reasonable.

markelliot:
@uschi2000, how's this?

uschi2000:
👍

PR #75, opened 2016-10-27
Add possibility to copy and preserve a folder. Also allow to copy the…

guenhter:
… content of a folder by using a '/*' at the very end

I had the problem, that gradle-tasks had output-files, which point to directories. 
But the gradle-copy task with `from ... into ...` not preservs that folder (https://docs.gradle.org/current/userguide/working_with_files.html#sec:copying_files) and just copies the content of it.

This PR allows:

```
task outputTask {
    outputs.files('some_folder', 'some_other_folder/*')
}
docker {
...
    dependsOn outputTask
    files 'my_folder', 'my_other_folder/*'
}
```

This will copy `my_folder`and `some_folder` as they are and the content of `my_other_folder` and `some_other_folder`.

So everything is still possible, what was possible before.

_Breaking Change:_
If someone relies that this plugin only copies the content of a folder, this must be re-configured by adding a `/*`.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
@@ -99,7 +99,7 @@ class DockerExtension {
         return resolvedDockerComposeFile
     }
 
-    public Set<String> getResolvedFiles() {
+    public Set<File> getResolvedFiles() {
         return resolvedFiles
     }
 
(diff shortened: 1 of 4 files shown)


guenhter:
Change the java8 method 'forEach' to a good old groovy 'each' method

guenhter:
Taking back pull request, because something is now broken with the dependencies.

jomach:
is there a way to use this ?

guenhter:
I didn't investigate any more on this issue. But as far as I remember, by putting a folder into a dummy folder woks. So all subfolders of the dummy folder are than copied.

jomach:
I end up using the distribution plug in. The docker add command unpacks the tar file automatically.
Works very good

for completion I have on my gradle file: 
distributions {
	main {
		contents {
			into('conf') {
				from rootProject.files("somefolder/conf")
			}
			from jar.outputs
		}
	}
}

docker {
	name imagename
	files tasks.distTar.outputs <- This does the trick
	dockerfile "dockerFile"
	tags "$artifactVersion"
}

I cannot format this sorry

guenhter:
Thanks for sharing this information. Sounds like a very nice workaround/alternative.

Issue #81, opened 2016-11-29
Docker-compose resolves dependencies during configuration time


henryptung:
While `generateDockerCompose` does need to use the dependencies from the `docker` configuration for further operation, I think it can do so during the task execution stage rather than project configuration stage. This accelerates project evaluation during the (common) case where `generateDockerCompose` is not being executed.

uschi2000:
Yep, agreed. Don you want to give it a shot?

stijndehaes:
This can get closed?

PR #82, opened 2016-12-02
Avoid configuration resolution during project configuration

henryptung:
Targets #81.

- Defer resolution until task up-to-date check
- Avoid downloading artifacts during resolution (e.g. large dists)
- Avoid using afterEvaluate

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/GenerateDockerCompose.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/GenerateDockerCompose.groovy
@@ -17,6 +17,10 @@ class GenerateDockerCompose extends DefaultTask {
     DockerComposeExtension ext
     Configuration configuration
 
+    GenerateDockerCompose() {
+        group = 'Docker'
+    }
+
     @TaskAction
     void run() {
         if (!template.file) {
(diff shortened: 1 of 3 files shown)


henryptung:
Of note: this changes the task to no longer resolve the input files at a later date; instead, it just resolves files in-place (since the conversion from e.g. `String` -> `File` should be deterministic unless someone, say, changes the path of the Gradle project mid-evaluation :P ).

src/main/groovy/com/palantir/gradle/docker/GenerateDockerCompose.groovy
```diff
@@ -0,0 +1,78 @@
...
+                " of dependencies from the '${configuration.name}' configuration"
+        return super.description ?: defaultDescription
+    }
+
+    @Input
+    @Memoized
+    Set<ModuleVersionIdentifier> getModuleDependencies() {
+        log.info "Resolving Docker template dependencies from configuration ${configuration.name}..."
+        return configuration.resolvedConfiguration
+            .firstLevelModuleDependencies
```
uschi2000: this looks like a behaviour change; would prefer to introduce that in a separate PR

uschi2000 (commented):
thanks, this is much better. one question on the possible behaviour change, otherwise LGTM

henryptung:
@uschi2000 Thanks for the review; issues should be fixed now, and added a task group onto `GenerateDockerCompose` so it looks a little better.

uschi2000:
lgtm. Mark?

markelliot:
fine with me if it's fine with you

PR #84, opened 2016-12-08
Add 'labels' parameter for building images

andybradshaw:


src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -417,6 +417,31 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
...
+                name '${id}'
+                labels(['test_label': 'test_value'])
+            }
+        """.stripIndent()
+        when:
+        BuildResult buildResult = with('docker').build()
+
+        then:
+        buildResult.task(':docker').outcome == TaskOutcome.SUCCESS
+        exec("docker inspect --format '{{.Config.Labels}}' ${id}").contains("test_label")
```
andybradshaw: Interestingly, this gives me errors when trying to specify `json` in the format, even though I can run the following locally if I don't remove the image after building it:

```
$ docker inspect --format '{{json .Config.Labels}}' id10
{"test_label":"test_value"}
```

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -103,6 +103,11 @@ class PalantirDockerPlugin implements Plugin<Project> {
                     buildCommandLine.addAll('--build-arg', "${buildArg.getKey()}=${buildArg.getValue()}")
                 }
             }
+            if (!ext.labels.isEmpty()) {
+                for (Map.Entry<String, String> label : ext.labels.entrySet()) {
+                    buildCommandLine.addAll('--label', "${label.getKey()}=${label.getValue()}")
```
markelliot: Should likely do some string validation on the key, and wondering if the value needs to be quoted.

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -417,6 +417,31 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
...
+            FROM alpine:3.2
+        """.stripIndent()
+        buildFile << """
+            plugins {
+                id 'com.palantir.docker'
+            }
+
+            docker {
+                name '${id}'
+                labels(['test_label': 'test_value'])
```
markelliot: Should be possible to apply this without the parens or square brackets, I think.

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
markelliot: Possibly because values are not being quoted as strings?

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
andybradshaw: Yeah, that's a good call. Should we try to enforce any of the bullet points [on the docker label recommendations]( https://docs.docker.com/engine/userguide/labels-custom-metadata/#/key-format-recommendations)?

andybradshaw: Quoting the value leads to those quotes getting escaped and wrapped in with their own quotes in the labels that get stored:

```
$ docker inspect --format '{{json .Config.Labels}}' id10-quoted
{"another.label":"\"another.value\"","test-label":"\"test-value\""}

$ docker inspect --format '{{json .Config.Labels}}' id10-unquoted
{"another.label":"another.value","test-label":"test-value"}

$ docker inspect --format '{{json .Config.Labels}}' test-manifest-quoted
{"manifest": "\"{\"manifest-version\":\"1.0\",\"product-type\":\"service.v1\",...}\""}
```

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
andybradshaw: I don't think that's the issue. After quoting the values, if I print the output of the `exec` from the tests I get:

```
println exec("docker inspect --format '{{json .Config.Labels}}' ${id}")
[no output]

println exec("docker inspect --format '{{.Config.Labels}}' ${id}")
'map[another.label:"another.value" test-label:"test-value"]'
```

compared to doing the same thing with the docker cli from a shell:

```
$ docker inspect --format '{{json .Config.Labels}}' id10
{"another.label":"\"another.value\"","test-label":"\"test-value\""}

$ docker inspect --format '{{.Config.Labels}}' id10
map[another.label:"another.value" test-label:"test-value"]
```

I don't know if it's worth investigating further; the normal format appears to just be using golang's `%v` to print the map, so we can still check it for correctness.

andybradshaw:
Build failures are due to the version of docker not having the `--label` flag in [1.10.0](https://docs.docker.com/v1.10/engine/reference/commandline/build/). It was added in [1.11.0](https://docs.docker.com/v1.11/engine/reference/commandline/build/), but I don't know if we prefer small or large bumps when it comes to docker.

Edit: a lot of people have requested [support for newer docker versions](https://discuss.circleci.com/t/support-docker-versions-newer-than-v1-9-1/7789/1) in circle but it looks like until the circle 2.0 update (in closed beta now), we're probably stuck on 1.10 ☹️

uschi2000:
I think believe we should hold off of merging this until the required Docker version is widely available.

markelliot:
Pretty sure 1.11 is widely available. Should just fix the circle setup to use it.

andybradshaw:
Per [this comment thread](https://discuss.circleci.com/t/support-docker-versions-newer-than-v1-9-1/7789/1), I don't believe Circle has docker 1.11 working, but if you know of people who have it working, I'd be glad to try to bump to it.

We can request access to the Circle 2.0 beta, but I've not seen any concrete information about what they are doing to have better docker support.

andybradshaw:
Wanted to check here to see if we have a path forward. Would like to get this change in sometime soon.

gdearment:
@andybradshaw - where does this stand?

uschi2000:
@andybradshaw can you please make the upgrade-circle thing a separate PR?

uschi2000:
also, this needs a rebase

andybradshaw:
@uschi2000 Circle 2 should be ready to go, and I've rebased... ready for another pass.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -94,6 +98,15 @@ class PalantirDockerPlugin implements Plugin<Project> {
                     buildCommandLine.addAll('--build-arg', "${buildArg.getKey()}=${buildArg.getValue()}")
                 }
             }
+            if (!ext.labels.isEmpty()) {
+                for (Map.Entry<String, String> label : ext.labels.entrySet()) {
+                    if (!label.getKey().matches(labelKeyPattern)) {
+                        throw new GradleException(String.format("Docker label '%s' contains illegal characters. " +
+                                "Label keys must only contain lowercase alphanumberic, `.`, or `-` characters.", label.getKey()))
```
uschi2000: prefer defining the pattern as a string so can log the explicit expected pattern here.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -29,9 +30,12 @@ import org.gradle.api.tasks.Delete
 import org.gradle.api.tasks.Exec
 import org.gradle.api.tasks.bundling.Zip
 
+import java.util.regex.Pattern
+
 class PalantirDockerPlugin implements Plugin<Project> {
 
     private static final Logger log = Logging.getLogger(PalantirDockerPlugin.class)
+    private static final Pattern labelKeyPattern = Pattern.compile("^[a-z0-9.-]*\$")
```
uschi2000: maybe CONST_CASE?

uschi2000: I'm a bit surprised by that `\$`, why is it escaped?

andybradshaw: Without escaping it, it fails to compile with `illegal string body character after dollar sign`, which I believe is due to Groovy string replacement. Replacing the double quotes with single quotes fixes this.

uschi2000 (approved):

PR #89, opened 2016-12-11
Fix #86: Exclude "buildDir/docker" instead of "buildDir" and fix infinite loop when explicitly including everything

### LINKED ISSUE #86: ADD build/distributions/*.tar (retain folder structure using files or include from project.buildDir?)
Hi, I really like the approach of this plugin but have encountered a problem which prohibits me form using it:

I have an artifact that is created by `task:distTar` and is located in `build/distributions`. From viewing the source code I can see that when no `files` argument is provided, the plugin will include all files _except_ the ones in the `build` directory. This becomes problematic because my artifact is actually located in the `build` folder. And when I try to _explicitly_ include my artifact, the path is not retained and the file is instead placed in the root of the `build/docker` sub-directory.

This means that I would either have to relocate my artifact manually before running this plugin, or change my Dockerfile to look for the file in the root instead of in the build folder, neither of which I want to do.

I have two suggestions to solve this problem:

1. optionally keep the folder structure when explicitly including files using the `files` argument. (preferred)
2. optionally include files from `build` (good enough)

(I also tried to explicitly include "everything" but that seems to fail due to a stack overflow when creating the path, probably a bug. Steps to reproduce: `files "${projectDir}"` or `files '.'`).

Thanks a lot!

/Johannes

jsundqvist:
Hey guys,

I know you want new features/fixes to be properly tested but I do not want to start refactoring a 120 line method that has 0 test coverage. I will instead document the test cases here so you can hopefully include them in your test suite once you decide to make `PalantirDockerPlugin.apply` testable.

test1: should include `buildDir` in `buildDir/docker`
```
docker {
	name 'myorg/myapp'
	dependsOn distTar
}
```

test2: should not crash
```
docker {
	name 'myorg/myapp'
	files '.' // or "${projectDir}"
	dependsOn distTar
}
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,23 +42,34 @@ if (System.env.CIRCLE_TEST_REPORTS) {
 }
 
 group 'com.palantir.gradle.docker'
-version gitVersion()
+version '0.9.3'
 
 task sourceJar(type: Jar) {
     from sourceSets.main.allSource
     classifier 'sources'
 }
-  
+
 publishing {
     publications {
-        bintray(MavenPublication) {
+        mavenJava(MavenPublication) {
             from components.java
-            artifact(sourceJar)
-            artifact(publishPluginJavaDocsJar)
         }
     }
 }
-  
+
+publishing {
+    repositories {
+        mavenLocal()
+        maven {
+            url "http://israproductions.com:8081/repository/maven-releases"
+            credentials {
+                username "admin"
+                password "admin123"
+            }
+        }
+    }
+}
+
 bintray {
     user = System.env.BINTRAY_USER
     key = System.env.BINTRAY_KEY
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @jsundqvist! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

uschi2000:
fyi, i added a test for the dependsOn functionality here: https://github.com/palantir/gradle-docker/pull/91

uschi2000:
agree that Plugin#apply is still not all that easy to test :(

jsundqvist:
Okey, I had a look at your test and tried to create new ones that reflect my use case and the expected behavior.

I guess the thing is that I want to be able to use the official distribution plugin and its predefined distTar task without having to change destinationDir.

The tests in #92 fail without the fix but pass in here.

uschi2000:
superseded by #101

PR #92, opened 2016-12-13
Verify issue 86

jsundqvist:


src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -417,6 +417,65 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
         execCond("docker rmi -f ${id}") || true
     }
 
+    def 'outputs from default distribution locations are added to the docker context'() {
```
uschi2000, line 420: I don't think this is expected functionality.

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -417,6 +417,65 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
...
+        when:
+        BuildResult buildResult = with('docker').build()
+
+        then:
+        buildResult.task(':dockerPrepare').outcome == TaskOutcome.SUCCESS
+        buildResult.task(':docker').outcome == TaskOutcome.SUCCESS
+        execCond("docker rmi -f ${id}") || true
+    }
+
+    def 'explicitly adding everything does not crash due to recursive overflow'() {
```
uschi2000, line 449: OK, this is a bug.

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
jsundqvist: So you are saying I cannot use the default artifact location of the official Gradle distribution plugin?  Then I have to either hack my Dockerfile or create a custom tar task for your plugin to even work.

uschi2000: You can use it by declaring `dependsOn distTar`.

jsundqvist: I can't believe your unwillingness to see the problem. Hopefully this will make it clear:
https://circleci.com/gh/jsundqvist/palantir-gradle-docker-issue-86/4
https://github.com/jsundqvist/palantir-gradle-docker-issue-86

The problem is that a standard Dockerfile expects the artifact to be located in `build/distributions/` when using the official distribution plugin and its default artifact location. The artifact that you copy when using `dependsOn distTar` is placed in the root, so the Dockerfile cannot find it.

The change I have provided solves the problem in one way, but maybe it's not the path you want to go. Either way I have decided not to use your plugin since this has all been way to much of a hassle.

Good luck with you codebase.

uschi2000: Johannes, I am trying to understand your problem. Here is a test that suggests what I have been proposing from the start of this discussion: it declares a dependency on a standard Gradle `distTar` task and demonstrates that one can build a Docker image from the resulting tar file: https://github.com/palantir/gradle-docker/pull/93

The Docker image built here contains the contents of the tar file as expected per the Dockerfile:
```
can-build-Docker-image-from-standard-Gradle-distribution-plugin_6
/tmp # cd can-build-Docker-image-from-standard-Gradle-distribution-plugin_6/
/tmp/can-build-Docker-image-from-standard-Gradle-distribution-plugin_6 # ls
bin  lib
/tmp/can-build-Docker-image-from-standard-Gradle-distribution-plugin_6 # find .
.
./lib
./lib/can-build-Docker-image-from-standard-Gradle-distribution-plugin_6.jar
./bin
./bin/can-build-Docker-image-from-standard-Gradle-distribution-plugin_6
./bin/can-build-Docker-image-from-standard-Gradle-distribution-plugin_6.bat
/tmp/can-build-Docker-image-from-standard-Gradle-distribution-plugin_6 #
```

Please let me know if this clarifies the usage. I agree that the plugin code (including tests and to some extent documentation) have grown a bit out of control over the past months and we're happy to fix bugs and make improvements where needed. As such, we have indeed had good luck with the plugin and have been using it happily.

uschi2000:
superseded by #101

Issue #94, opened 2016-12-20
Ability to easily push all tags for an image


rmelick:
It would be great if we could easily push all tags for an image.  The dockerTag task is nice for tagging them all, but currently dockerPush only pushes the main named image, and does not push those other tags.

devkanro:
See this PR #205 , and use `dockerTagPush` task to push taged image only, `dockerAllPush` task to push all

PR #95, opened 2016-12-20
Make push task depend on push tasks for each tag if tags are present

rmelick:
Issue #94 : Now, dockerPush will depend on all the dockerPushTagName tasks, which in turn depend on the dockerTagTagName tasks.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
@@ -132,13 +132,14 @@ class PalantirDockerPlugin implements Plugin<Project> {
                     })
                     tag.dependsOn subTask
 
-                    project.tasks.create('dockerPush' + taskTagName, Exec, {
+                    Exec pushSubTask = project.tasks.create('dockerPush' + taskTagName, Exec, {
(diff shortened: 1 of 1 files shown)


uschi2000:
I'm not convinced this is desirable behavior: it makes it pretty hard to push tags selectively in CI scripts, no?

rmelick:
It does make it difficult to push tags selectively.  Put, there is also no good way to push all the tags.  Do you think it would be better to have a task that pushed all tags (like dockerPushAll)?  That way you could also push selectively if desired?

pkoenig10:
The `dockerPush` task runs `docker push <NAME>` which pushes all tags.

devkanro:
See this PR #205 :

Use `dockerAllPush` can solve this issue/PR without change behavior of `dockerPush`, use `dockerTagsPush` can only push the tags.

### New tasks in PR:
**dockerTagsPush**
Pushes all tagged Docker images to configured Docker Hub.
```
dockerTagsPush= dockerPush<Tag1>+ dockerPush<Tag2> + ... + dockerPush<Tagn>
```

**dockerAllPush**
Pushes all tagged Docker images and named Docker image to configured Docker Hub.
```
dockerAllPush = dockerTagsPush + dockerPush
```

PR #101, opened 2017-01-24
Remove dependsOn configuration, use CopySpec to evaluate `files`

uschi2000:


uschi2000:
@markelliot, this is a breaking change, but one I believe we should make.

markelliot:
yeah, it's fine. I guess being idiomatic is going to beat convenience (ie most of the time you want not just to consume the distTar outputs but also to depend on the actual task.

uschi2000:
you can still do exactly that, but with a few more characters. it's more gradle standard style, i guess.
```
distribution {
  files tasks.distTar.outputs
}
```

instead of
```
distribution {
  dependsOn distTar
}
```

markelliot:
Does saying you depend on tasks.distTar.outputs actually mean you depend on distTar? (at least in Gradle 2.x this was not the case)

uschi2000:
yes it does. there's a test for this (see latest commit).

markelliot:
wfm 👍

PR #111, opened 2017-02-14
Update circle.yml file to circle version 2

andybradshaw:


uschi2000:
needs to fix creds, spoke offline

circle.yml
```diff
@@ -1,26 +1,20 @@
...
-  artifacts:
-    - build/libs
-
-test:
-  override:
-    - ./gradlew --parallel -i build
-
-deployment:
-  release:
-    tag: /[0-9]+(\.[0-9]+)+((-(beta|rc)[0-9]{1,2})(\.[0-9])?)?/
```
uschi2000, line 23: need to ensure that this logic is carried over.

uschi2000: it's fine to do this in gradle rather than circle

circle.yml
```diff
@@ -1,26 +1,20 @@
...
-    commands:
-      - ./gradlew -i bintrayUpload
-      - ./gradlew -i -Dgradle.publish.key=$GRADLE_KEY -Dgradle.publish.secret=$GRADLE_SECRET publishPlugins
+version: 2
+executorType: machine
+stages:
+  build:
+    workDir: ~/gradle-docker
```
uschi2000, line 5: this is not the default? i'm surprised.

uschi2000:
Any idea why the test results XML doesn't show up in Circle?

uschi2000 (approved):

andybradshaw:
Not sure... looks like this happened on Mark's Circle 2.0 PR as well. Will play around with it on a branch and see what's up.

Edit: This is fixed by #113

PR #112, opened 2017-02-14
Fix DockerRun test on Linux

andybradshaw:


uschi2000:
what breaks here, exactly?

andybradshaw:
Failure output:
```
    Output:
    FAILURE: Build failed with an exception.
    * What went wrong:
    A problem occurred configuring root project 'can-mount-volumes_5'.
    > dockerfile 'Dockerfile' does not exist.
```

If the Dockerfile is written to the 'test' directory, there is no Dockerfile in the top level of the context sent to the daemon.

uschi2000:
I don't understand how that test every worked... 
Can you also remove the isLinux thing? (This works now on MacOS)

PR #113, opened 2017-02-14
Update circle.yml to store test results

andybradshaw:
Need to make a PR to test this.


andybradshaw:
@uschi2000 this should fix the issue with storing the test results

uschi2000:
I don't see the test results in circle; do they only show up when tests fail?

PR #121, opened 2017-03-01
Adding a test which checks if a directory formatted file gets preserved

aclairekeum:
When a directory is added to 'files' it gets unpacked when dockerPrepare. This test is to check that the directory is still preserved as it is.


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @aclairekeum! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

uschi2000:
It's easier to solve this use-case with the underlying CopySpec, see here for an example: https://github.com/palantir/gradle-docker/pull/123

PR #130, opened 2017-03-31
Make dockerfile a file and skip resoultion if overridden

andybradshaw:


src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
```diff
@@ -27,7 +27,8 @@ class DockerExtension {
     Project project
 
     private String name = null
-    private String dockerfile = 'Dockerfile'
+    private String dockerfilePath = 'Dockerfile'
```
uschi2000: should be a constant

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
```diff
@@ -118,8 +119,13 @@ class DockerExtension {
         Preconditions.checkArgument(!Strings.isNullOrEmpty(name),
             "name is a required docker configuration item.")
 
-        resolvedDockerfile = project.file(dockerfile)
-        Preconditions.checkArgument(resolvedDockerfile.exists(), "dockerfile '%s' does not exist.", dockerfile)
+        if (dockerfile != null) {
+            resolvedDockerfile = dockerfile
+        } else {
+            resolvedDockerfile = project.file(dockerfilePath)
+            Preconditions.checkArgument(resolvedDockerfile.exists(),
```
uschi2000: this check should run in both cases

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -56,25 +56,6 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
         buildResult.output.contains("name is a required docker configuration item.")
     }
 
-    def 'fail with missing dockerfile'() {
```
uschi2000, line 59: should keep this test, right?

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -105,7 +86,7 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
     def 'check plugin creates a docker container with non-standard Dockerfile name'() {
         given:
         String id = 'id2'
-        file('foo') << """
+        File foo = file('foo') << """
```
uschi2000: don't need this change

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -116,7 +97,7 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
 
             docker {
                 name '${id}'
-                dockerfile 'foo'
+                dockerfile new File('foo')
```
uschi2000: should probably use `project.file("foo")` or similar

uschi2000 (changes requested):
also need to change the readme

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
andybradshaw: I agree, but I think moving validation of the paths to task execution is enough work that it isn't worth dealing with it in this PR.  The docker daemon will throw a useful enough error here to let someone with a misconfigured plugin know what has gone wrong, which the purpose of this error.

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
uschi2000: ping. why is this test gone?

andybradshaw: I removed the test since we're no longer the ones throwing the error in this case, docker is. We can still have this test, but it would only be testing the failure of the task and not the error that we used to throw.

readme.md
```diff
@@ -24,16 +24,16 @@ plugins {
...
 automatically include outputs of task dependencies in the Docker build context.
 
 **Docker Configuration Parameters**
 - `name` the name to use for this container, may include a tag
 - `tags` (optional) an argument list of tags to create; any tag in `name` will
   be stripped before applying a specific tag; defaults to the empty set
-- `dockerfile` (optional) dockerfile to use for building the image; defaults to
-  `${projectDir}/Dockerfile`
+- `dockerfile` (optional) the dockerfile to use for building the image; defaults to
+  reading the file at the path `${projectDir}/Dockerfile`
```
uschi2000: should probably say "defaults to `project.file('Dockerfile')`" and say that this must be a file object.

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
```diff
@@ -118,8 +119,13 @@ class DockerExtension {
         Preconditions.checkArgument(!Strings.isNullOrEmpty(name),
             "name is a required docker configuration item.")
 
-        resolvedDockerfile = project.file(dockerfile)
-        Preconditions.checkArgument(resolvedDockerfile.exists(), "dockerfile '%s' does not exist.", dockerfile)
+        if (dockerfile != null) {
+            resolvedDockerfile = dockerfile
+        } else {
+            resolvedDockerfile = project.file(DEFAULT_DOCKERFILE_PATH)
+            Preconditions.checkArgument(resolvedDockerfile.exists(),
```
uschi2000: this check should run either never or in both cases.

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
uschi2000: OK. i think we need to make the behavior more symmetric, see note above.

uschi2000 (approved):

PR #132, opened 2017-04-20
Change usage of dockerfile to file-object in documentation

guenhter:
Documentation was misleading when example is used as it is


uschi2000 (approved):
Thanks, I missed this one.

PR #137, opened 2017-06-08
Specify Source & Target Compatibility (set to 1.7)

sshankar:
Published plugin artifact inherits it's source and target java compatibility version, 1.8, from the build environment. This PR pins `sourceCompatibility` and `targetCompatibility` to the minimum possible version, 1.7, expanding the plugin's usage scope.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,6 +30,9 @@ if (System.env.CIRCLE_TEST_REPORTS) {
 group 'com.palantir.gradle.docker'
 version System.env.CIRCLE_TAG ?: gitVersion()
 
+sourceCompatibility = 1.7
+targetCompatibility = 1.7
+
 task sourceJar(type: Jar) {
     from sourceSets.main.allSource
     classifier 'sources'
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @sshankar! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

build.gradle
```diff
@@ -30,6 +30,9 @@ if (System.env.CIRCLE_TEST_REPORTS) {
 group 'com.palantir.gradle.docker'
 version System.env.CIRCLE_TAG ?: gitVersion()
 
+sourceCompatibility = 1.7
+targetCompatibility = 1.7
```
uschi2000: i believe that targetC inherits from sourceC? if that is true, the first line should suffice.

sshankar: Yes, but there's no harm in specifying it.

uschi2000: ok, let's do sourceC only, please. (since we usually only specify that in our projects.)

uschi2000:
thanks!

PR #142, opened 2017-08-08
Docker plugin: added 2 features - 1) ability to add hosts to /etc/hosts 2) ability to link other containers

nminyashkin:


nminyashkin:
Hi there, will someone have a look at this? Thanks!

uschi2000:
docker-compose provides these features. why is this necessary?

pkoenig10:
Container links have been deprecated:
https://docs.docker.com/network/links/

Networking between containers should be done with user-defined bridge networks. This can be easily done with docker-compose which will create a single bridge network connecting all the services defined in your compose file.

snagafritz:
`--add-host` is a feature of both `docker build` and `docker run`. would be nice if it were also supported by this plugin. I have a case right now where I need custom `/etc/hosts` entries in my container in order to communicate with an external machine.

PR #147, opened 2017-10-18
Resolve volume locations using project.file() to support absolute and relative paths (fixes #138)

### LINKED ISSUE #138: Volumes key in dockerRun should not be relative to project directory
As the readme states:

> volumes optional map of volumes to mount in the container. The key is path to the host volume, relative to the project folder, the value is the exposed container volume path.

The keys in the volumes property of the dockerRun task are relative to the project directory. This doesn't make sense for mounting things like `.pgpass` files into a docker container that you don't want to check in to your repository.

The keys in the volumes property should be absolute paths, and if somebody wants them to be relative they can use `${rootDir}` or `${projectDir}`

tmccarthy:
Currently, the host path for volumes is always resolved relative to the project directory. This means that for a project at `/Users/timothy/gradle-docker`, the volume path `/tmp/volume` would be resolved as `/Users/timothy/gradle-docker/tmp/volume`. This is counterintuitive.

In this pull request, the host volume is resolved using the more idiomatic [`project.file()`](https://docs.gradle.org/current/userguide/working_with_files.html#sec:locating_files) method. Relative paths are still resolved relative to the project directory, but absolute paths are now resolved as expected. Additionally, volumes can now be specified as either strings or `File` objects.

A summary of changed behaviour (for a project at `/Users/timothy/gradle-docker`) is below:

| Specified volume path | Path used currently | Path used in PR |
|---|---|---|
| `"volume"` | `/Users/timothy/gradle-docker/volume` | `/Users/timothy/gradle-docker/volume` |
| `"/tmp/volume"` | `/Users/timothy/gradle-docker/tmp/volume` | `/tmp/volume` |
| `new File("volume")` | Error | `/Users/timothy/gradle-docker/volume` |
| `new File("/tmp/volume")` | Error | `/tmp/volume` |

This pull request fixes #138.


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @tmccarthy! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/test/groovy/com/palantir/gradle/docker/DockerRunPluginTests.groovy
```diff
@@ -201,6 +201,53 @@ class DockerRunPluginTests extends AbstractPluginTest {
         buildResult.output =~ /(?m):dockerRunStatus\nDocker container 'foo' is STOPPED./
     }
 
+    def 'can mount volumes specified by a File object with an absolute path'() {
```
pkoenig10: I would prefer we change this test to just "can mount volumes specified with an absolute path" and use the absolute path string in the build file. While using a File object works, its a bit awkward and I would rather not endorse that usage in the tests.

tmccarthy: Yep, I've updated the PR. Thanks!

src/main/groovy/com/palantir/gradle/docker/DockerRunExtension.groovy
```diff
@@ -30,7 +30,7 @@ class DockerRunExtension {
     private List<String> command = ImmutableList.of()
     private Set<String> ports = ImmutableSet.of()
     private Map<String,String> env = ImmutableMap.of()
-    private Map<String,String> volumes = ImmutableMap.of()
+    private Map<Object,String> volumes = ImmutableMap.of()
```
pkoenig10, line 33: Actually would prefer to just leave this as a string also. Not sure there is any value in supporting file objects.

uschi2000: The `project.file()` behavior is very much the standard in Gradle.

pkoenig10 (approved):

PR #163, opened 2018-02-14
Allow for configuration of additional template tokens for docker compose files.

darora:
Allows for automated configuration of what the current image under testing ought to be, thereby avoiding the requirement of manual configuration when images are published to snapshot repositories vs release repositories.

### DIFF
--- a/src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
@@ -41,7 +41,7 @@ class DockerComposePluginTests extends AbstractPluginTest {
             }
 
             dockerCompose {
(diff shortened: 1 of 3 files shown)


iamdanfox (commented):
All looks good, would just prefer to supply arbitrary keys & values instead of the hard-coded `currentImageName` :)

iamdanfox (approved):

src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
```diff
@@ -38,6 +40,10 @@ class DockerComposePluginTests extends AbstractPluginTest {
                 jcenter()
             }
 
+            dockerCompose {
+                templateTokens(['currentImageName': 'snapshot.docker.registry/current-service:1.0.0-1-gabcabcd'])
```
jmcampanini, line 44: no test for `templateToken(String key, String value)` :(

PR #167, opened 2018-02-16
Capture tagName in lazily evaluated block

darora:


iamdanfox (approved):
Sneaky freaking closures

Issue #162, opened 2018-01-22
Problem on mac: "Cannot run program "docker": error=2, No such file or directory"


solomkinmv:
When I try to build docker I receive following error message:
`Starting process 'command 'docker''. Working directory: /Users/msolomkin/Projects/glossary/discovery-server/build/docker Command: docker build --build-arg JAR_FILE=discovery-server-0.0.1-SNAPSHOT.jar -t solomkinmv/discovery-server .`

But when I go to the working directory (`/Users/msolomkin/Projects/glossary/discovery-server/build/docker`) manually and run command `docker build --build-arg JAR_FILE=discovery-server-0.0.1-SNAPSHOT.jar -t solomkinmv/discovery-server .` everything works fine.

Here is the stack trace:

> FAILURE: Build failed with an exception.
> 
> * What went wrong:
> Execution failed for task ':discovery-server:docker'.
> > A problem occurred starting process 'command 'docker''
> 
> * Try:
> Run with --debug option to get more log output. Run with --scan to get full insights.
> 
> * Exception is:
> org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':discovery-server:docker'.
> 	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:100)
> 	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:70)
> 	at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
> 	at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:62)
> 	at org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter.execute(ResolveTaskOutputCachingStateExecuter.java:54)
> 	at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:60)
> 	at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:97)
> 	at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:87)
[...]

PR #172, opened 2018-04-02
Allow Docker binary location to be overridden by specifying environment variable GRADLE_DOCKER_BINARY

rikbrown:
## Notes

Allow Docker binary location to be overriden by specifying environment variable `GRADLE_DOCKER_BINARY`.

#### Justification

Not all build environments have `docker` in the path available during Gradle build.  For example, on Mac Java's `ProcessBuilder` (used by Gradle `Exec`) only considers `/usr/bin` and not `/usr/local/bin`, despite any `PATH` environment variables (see [this issue](https://github.com/palantir/gradle-docker/issues/162)).  This change allows builders to explicitly provide their docker location (e.g. `/usr/local/bin/docker`).  On my system, Docker was installed using Homebrew which put it in local bin and SIP prevents symlinking into bin.

#### Environment variable vs parameter in buildfile

I preferred an environment variable instead of providing `dockerBinary` option in the buildfile (like one other plugin does) to make this more portable.  Consumers of a library using this plugin should not be forced to use a specific binary path (definitely breaks in a cross-platform world too).  Instead, they can set this environment variable once and things will just work.

## Testing

* `gradle-docker$ gradle build`
  * Existing unit tests. These assume the default `docker` and keep working on platforms where that works. However, I modified them to also use the environment variable, which allowed them to pass on my system.
  * Added a new unit test which overrides the behaviour of getting an environment variable and tests the `DockerExecutor` trait directly to prove it uses the env but falls back to `docker`.  I chose to do this because you cannot trivially (and in a well-supported cross-platform manner) modify environment variables.
* `my-own-gradle-project$ GRADLE_DOCKER_BINARY=/usr/local/bin/docker gradle docker`
 * The Gradle project I'm working on which uses this plugin now finally works when the variable is set!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -52,7 +52,11 @@ docker {
 To build a docker container, run the `docker` task. To push that container to a
 docker repository, run the `dockerPush` task.
 
-Tag and Push tasks for each tag will be generated for each provided `tags` entry. 
+Tag and Push tasks for each tag will be generated for each provided `tags` entry.
+
+**Docker Environment Variables**
+- `GRADLE_DOCKER_BINARY` the path to the Docker binary to use; defaults to `docker` (and relies on any operating system
+  specific behaviour to resolve it, which [sometimes should not be relied upon](https://github.com/palantir/gradle-docker/issues/162)). 
 
 **Examples**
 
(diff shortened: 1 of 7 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @rikbrown! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

pkoenig10:
Locating the docker binary is an environment/setup issue and I don't think this plugin should be the solution to a misconfigured environment.

@rikbrown I have responded in #162 and would be happy to iterate further there.

PR #173, opened 2018-04-02
Change default template to docker-compose.template.yml

### LINKED ISSUE #71: Change template name to docker-compose.template.yml
Could we rename the `docker-compose.yml.template` template default filename to `docker-compose.template.yml`?

IDEs won't do yaml syntax highlighting on a yaml file with a ".template" extension.

I know I can use the following block:

``` groovy
dockerCompose {
    template 'docker-compose.template.yml'
}
```

But, I would prefer a better default to avoid adding these lines to my build.gradle

pkoenig10:
Fixes #71

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -138,7 +138,7 @@ those dependencies.
 ### Generating docker-compose.yml files from dependencies
 
 The `com.palantir.docker-compose` plugin uses the transitive dependencies of the
-`docker` configuration to populate a `docker-compose.yml.template` file with the
+`docker` configuration to populate a `docker-compose.template.yml` file with the
 image versions specified by this project and all its transitive dependencies.
 The plugin uses standard Maven/Ivy machanism for declaring and resolving
 dependencies.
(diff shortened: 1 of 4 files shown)


iamdanfox:
@pkoenig10 btw as you've been maintaining this repo without really needing the infra team's involvement, would you like us to get rid of that CODEOWNERs thing?

pkoenig10:
I'm good either way. It's helpful just to get another set of eyes on any changes I make.

iamdanfox:
Cool, for this specific PR, I think this would break a bunch of internal repo's usage.  Could we coordinate this with an excavator PR or even better, make this fallback to reading from the old file (just printing a warning)?

pkoenig10:
Yes but it will be a very obvious break and a simple fix.

Let's coordinate with an excavator PR. Falling back to another file would introduce a lot of complexity here.

iamdanfox:
@pkoenig10 did you look into the excavator approach here?

pkoenig10:
I haven't had the time, but this change is not high priority. I'll follow up here once I've looked into it.

PR #174, opened 2018-04-10
Add support for getting a dynamic host port

### LINKED ISSUE #171: Port of 0 is not allowed but is valid to docker
If a port of 0 is passed in the mappings then docker allocates a free port, but trying to use 0 in dockerRun results in error 'Port must be in the range [1,65536]'.

(As an aside, this: https://stackoverflow.com/questions/42482278/bash-command-to-return-a-free-port suggests how to find which port has been allocated)

pkoenig10:
Fixes #171

Ultimately I think it would better to stop using the given port as the host port when only one value is specified and instead dynamically allocate a host port to match the `docker run` behavior.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/DockerRunExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerRunExtension.groovy
@@ -107,11 +107,11 @@ class DockerRunExtension {
         for (String port : ports) {
             String[] mapping = port.split(':', 2)
             if (mapping.length == 1) {
-                checkPortIsValid(mapping[0])
+                checkContainerPortIsValid(mapping[0])
                 builder.add("${mapping[0]}:${mapping[0]}")
             } else {
-                checkPortIsValid(mapping[0])
-                checkPortIsValid(mapping[1])
+                checkHostPortIsValid(mapping[0])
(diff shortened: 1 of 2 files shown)


pkoenig10:
This should be kept open

pkoenig10:
This should be kept open

pkoenig10:
@palantir/infrastructure is someone able to review this PR?

dansanduleac:
If we're going to do this, then can we also expose the port it ends up using, as a [property](https://docs.gradle.org/4.4/javadoc/org/gradle/api/provider/Property.html) on the task (can use `ext.port` for now)?
I don't want people's build.gradles to start including tasks that run `docker inspect -f '{{ (index (index .NetworkSettings.Ports "????") 0).HostPort }}' <containerid or container name>`

```groovy
dockerRun.ext.port = project.objects.property(Integer)
dockerRun.ext.port.set(project.provider {
  // TODO figure out how to get containerPortMaybe by parsing out `ext.ports`,
  // or how to get the only mapping from `.NetworkSettings.Ports`
  exec {
    commandLine 'docker', 'inspect', """--format={{ (index (index .NetworkSettings.Ports "$containerPortMaybe") 0).HostPort }}""", ext.name
  }
})
```

Issue #149, opened 2017-10-24
dockerPush doesn't generate (and thus push) all Tags


pluttrell:
I have the following config:

```
docker {
    name 'my-company/my-image'
    tags project.version, 'latest'
    dockerfile file('src/main/docker/Dockerfile')
    dependsOn assemble
    files jar
}
```

If I remove all images from my local machine other then the base image and then do:

```
$ gradle clean dockerPush

> Task :rest-service:dockerPush
The push refers to a repository [docker.io/my-company/my-image]
07cb7e265fa0: Preparing
762429e05518: Preparing
2be465c0fdf6: Preparing
5bef08742407: Preparing
2be465c0fdf6: Layer already exists
5bef08742407: Layer already exists
762429e05518: Layer already exists
07cb7e265fa0: Pushed
latest: digest: sha256:09ae39c3379061413cc450148fdb32baa2b07616fed5341df542595b3eb08b5f size: 1154

BUILD SUCCESSFUL in 10s
11 actionable tasks: 8 executed, 3 up-to-date
```

Notice however that *only* an image with the `latest` tag is created, so naturally that was all that was pushed:

```
$ docker images
REPOSITORY                     TAG                 IMAGE ID            CREATED             SIZE
my-company/my-image            latest              c588bc51ad2e        7 seconds ago       81.4MB
openjdk                        8u131-jre-alpine    e2f6fe2dacef        5 weeks ago         81.4MB
```

To continue the example, I remove the image just created as such:

```
$ docker rmi c588bc51ad2e
Untagged: my-company/my-image:latest
Untagged: my-company/my-image@sha256:09ae39c3379061413cc450148fdb32baa2b07616fed5341df542595b3eb08b5f
Deleted: sha256:c588bc51ad2e5019176b09377fa9b15caa92216f6540dfb880c914dbafb69e47
Deleted: sha256:63d4d33b985b3ef47804d7e25cea23d91f4bef88a69168da2b54721aadd5eb97
Deleted: sha256:595c32029464585e6871fc32d075f12e4a175bae5c9d016c9fa9e6fc8975168c
Deleted: sha256:46e42747ee2fd820e49b90be14f066bdee60d589ca914eaf937df5aec811e037

$ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
openjdk             8u131-jre-alpine    e2f6fe2dacef        5 weeks ago         81.4MB
```

[...]

Issue #151, opened 2017-11-16
Unable to use dependsOn dockerTag with task dockerPush


rakshit-jain:
I have multiple tags in my Gradle build. Running gradle clean docker uses only one tag as mentioned in https://github.com/palantir/gradle-docker/issues/149. 

While applying the advice in the above issue I have encountered following error.
```
* What went wrong:
A problem occurred evaluating project ':test'.
> Could not get unknown property 'dockerTag' for project ':test' of type org.gradle.api.Project.
```
**Gradle Version**: 4.1

**Gradle Build**
```
docker{
    name myDocker
    tags version,'latest'
    dockerfile file('src/main/docker/Dockerfile')
    files jar, "src/main/docker/start.sh"
    buildArgs groupMap[System.getenv("ENVIRONMENT")]
    pull true
    noCache true
    dependsOn(build)
}

tasks.dockerPush.dependsOn(tasks.dockerTag)
```

uschi2000:
can you try `tasks.dockerPush.dependsOn(tasks.dockerTag)`?

rakshit-jain:
Error is same 
`> Could not get unknown property 'dockerTag' for task set of type org.gradle.api.internal.tasks.DefaultTaskContainer.`

I think this issue relates to https://github.com/palantir/gradle-docker/issues/139

joffrey-bion:
Hi, I got it to work with a closure, as suggested by @pluttrell in https://github.com/palantir/gradle-docker/issues/149:

    tasks.dockerPush.dependsOn { tasks.dockerTag }

PR #177, opened 2018-04-30
fix issue #149 and issue #151

thunderstumpges:
- (issue #151) Create the dockerTag task outside of the afterEvaluate so that it can be referenced during build configuration
- (issue #149) Make dockerPush depend on dockerTag
- make dockerPush\<tag\> only depend on dockerTag\<tag\>

### DIFF
--- a/src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
@@ -227,30 +227,6 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
         buildResult.task(':tasks').outcome == TaskOutcome.SUCCESS
     }
 
-    def 'no tag task when no tags defined'() {
-        given:
-        String id = 'id4'
-        file('Dockerfile') << """
-            FROM alpine:3.2
-            MAINTAINER ${id}
-        """.stripIndent()
-        buildFile << """
-            plugins {
-                id 'com.palantir.docker'
-            }
-
-            docker {
-                name '${id}'
-            }
-        """.stripIndent()
-
-        when:
-        BuildResult buildResult = with('tasks').build()
-
-        then:
-        !buildResult.output.contains('dockerTag')
-    }
-
     def 'tag and push tasks created for each tag'() {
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @thunderstumpges! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

thunderstumpges:
I see that a test "no tag task when no tags defined" failed. I don't see the harm in always having a dockerTag task (even if there's nothing to tag)... it allows us to write proper dependency chains in our own applications (fixing issue #151). I tried to remove the dockerTag task in the afterEvaluate, but it doesn't appear this is possible. I think we'll have to drop this one test if this PR is to make it in... thoughts?

reegnz:
@thunderstumpges it seems that you also need to fix a test-case because it assumes that the dockerTag target is not there if no tags are specified.

thunderstumpges:
Yeah, that's exactly my comment above. It's not that the test needs to be "fixed". It's that the test explicitly checks exactly what I removed. I think my only recourse is removing the test; I just wanted feedback if that was acceptable. I suppose I'll just commit that on the PR that way if anyone else wants it they can take it.

thunderstumpges:
hmm... well now I don't know what's going on. That test ('does not throw if name is configured after evaluation phase') passes on my machine.

thunderstumpges:
Okay... so I guess that test is intermittent... anyway, it's all "green" now. Please review!

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -314,6 +290,7 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
         execCond("docker rmi -f ${id}")
         execCond("docker rmi -f ${id}:another")
         execCond("docker rmi -f ${id}:latest")
+
```
pkoenig10, line 293: Remove extra line

pkoenig10 (approved):

PR #179, opened 2018-05-25
Update a few plugins in order to remove gradle depracation warning

guenhter:


build.gradle
```diff
@@ -17,7 +17,7 @@ repositories {
 
 dependencies {
     compile gradleApi()
-    compile 'com.google.guava:guava:18.0'
+    compile 'com.google.guava:guava:25.1-android'
```
pkoenig10: Would prefer not to upgrade guava

guenhter: Ok. Done.

pkoenig10 (approved):

PR #186, opened 2018-06-27
dockerComposeUp task

samrogerson:


src/main/groovy/com/palantir/gradle/docker/DockerComposeUp.groovy
```diff
@@ -0,0 +1,36 @@
+package com.palantir.gradle.docker
+
+import groovy.util.logging.Slf4j
+import org.gradle.api.DefaultTask
+import org.gradle.api.artifacts.Configuration
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.TaskAction
+
+@Slf4j
+class DockerComposeUp extends DefaultTask {
```
iamdanfox, line 10: Could we extend from an [Exec](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Exec.html) task so that people can configure/override more stuff?

src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
```diff
@@ -137,4 +137,41 @@ class DockerComposePluginTests extends AbstractPluginTest {
...
+        '''.stripIndent()
+        buildFile << '''
+            plugins {
+                id 'com.palantir.docker-compose'
+            }
+        '''.stripIndent()
+        when:
+        with('dockerComposeUp').build()
+        then:
+        file("foobarbaz").exists()
```
iamdanfox, line 175: :+1:

src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
```diff
@@ -137,4 +137,41 @@ class DockerComposePluginTests extends AbstractPluginTest {
         then:
         buildResult.output.contains("Could not find specified template file")
     }
+
+    def 'docker-compose is executed and fails on invalid file'() {
```
pkoenig10, line 141: Add a test specifiying a different docker-compose file

pkoenig10 (commented):
Can you also update the README with information about this task?

src/main/groovy/com/palantir/gradle/docker/DockerComposeUp.groovy
pkoenig10: Is there best practices regarding defining a task in a separate groovy class vs directly in the plugin class? We do the latter in `PalantirDockerPlugin` although a separate class makes things cleaner.

src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
```diff
@@ -137,4 +137,41 @@ class DockerComposePluginTests extends AbstractPluginTest {
...
+
+    def 'docker-compose successfully creates docker image'() {
+        given:
+        file('docker-compose.yml') << '''
+            version: "2"
+            services:
+              hello:
+                container_name: "helloworld"
+                image: "python:3.4-alpine"
+                command: touch /test/foobarbaz
```
pkoenig10: This can probably just be the alpine image (no python needed)

src/main/groovy/com/palantir/gradle/docker/DockerComposeUp.groovy
iamdanfox: Separate class is generally the way to go as it's much easier to add clear annotations for incremental behaviour, description etc etc.

It's also much easier to convert the separate classes from Groovy -> Java and gradle recommend writing plugins in Java over groovy https://docs.gradle.org/current/userguide/custom_plugins.html

samrogerson: Best practice is to not extend Exec.

src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
samrogerson: Yeah thanks for this - was infinitely better than what I'd done first.

src/main/groovy/com/palantir/gradle/docker/DockerComposeUp.groovy
iamdanfox: OK fair enough, this will probably be fine as an MVP anyway - can always add more configurability if people really need it later

uschi2000:
Directionally, I'd prefer to move this plugin back to towards its original intent: to create Docker images. Docker CLI wrapping seems like a bottomless endeavor, see some of the other issues on this repo.

src/main/groovy/com/palantir/gradle/docker/DockerComposeUp.groovy
samrogerson: Yeah - I would say composition with `project.exec` provides way more flexibility than directly extending `Exec` since you're bound to execs `TaskAction` method.

src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
samrogerson: done

src/test/groovy/com/palantir/gradle/docker/DockerComposePluginTests.groovy
samrogerson: done

samrogerson:
In agreement with Rob.  Longer term should break out the build orchestration utility tasks somewhere else.

pkoenig10:
Also in agreement about avoiding Docker CLI wrapping. The `dockerRun` plugin is the source of most of these feature creep issues.

Covnersely, I think adding Docker Compose tasks is valuable and less prone to feature creep. The `dockerCompose` plugin pushes most of the complexity and configurability into the provided template file. Adding some narrowly scoped tasks for managing Compose services allows us to keep the plugin simple while also supporting a wide range of workflows.

readme.md
```diff
@@ -152,6 +152,10 @@ replaces all occurrences of version variables of the form `{{group:name}}` in
 the `docker-compose.yml.template` file by the resolved versions and writes the
 resulting file as `docker-compose.yml`.
 
+The `docker-compose` plugin also provides a `dockerComposeUp` task that starts 
+the docker images specified in the `dockerComposeFile` in detached mode.
```
pkoenig10, line 156: Also add the `dockerComposeUp` task to the list of tasks at the bottom of the README

samrogerson: done

pkoenig10 (approved):

PR #198, opened 2018-09-03
Allow docker login to private repo before push

chris060986:
A new tasks is added which es executed before docker push, if a repository and credentials are defined in the dockerLogin section of the build.gradle.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
@@ -111,10 +111,16 @@ class PalantirDockerPlugin implements Plugin<Project> {
             clean.delete dockerDir
 
             login.with {
-                workingDir dockerDir
-                commandLine 'docker', 'login', '-u', "${ -> loginExt.username}", '-p', "${ -> loginExt.password}", "${ -> loginExt.repository}"
-                logging.captureStandardOutput LogLevel.INFO
-                logging.captureStandardError LogLevel.ERROR
+                if(loginExt.repository!=null) {
+                    workingDir dockerDir
+                    commandLine 'docker', 'login', '-u', "${-> loginExt.username}", '-p', "${-> loginExt.password}", "${-> loginExt.repository}"
+                    logging.captureStandardOutput LogLevel.INFO
+                    logging.captureStandardError LogLevel.ERROR
+                }
+                else {
+                    workingDir dockerDir
+                    commandLine 'echo', 'no login configured, skipping...'
+                }
             }
 
             prepare.with {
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @chris060986! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -153,6 +177,15 @@ class PalantirDockerPlugin implements Plugin<Project> {
         }
     }
 
+    private List<String> loginCommandline(DockerLoginExtension loginExt){
+        List<String> loginCommandLine = ['docker', 'login']
+        if(loginExt.username && loginExt.password){
+            loginCommandLine.addAll('-u', "${-> loginExt.username}", '-p', "${-> loginExt.password}")
```
iamdanfox: This isn't safe because if you run `./gradlew dockerLogin -i` (e.g. on CI) then the password will be logged to the console!

I ended up doing a `sh -c "docker login -u $USERNAME -p $PASSWORD"` where the values are passed in to the environment.

chris060986: I am not sure what you mean. Do you suggest the user should set the values $USERNAME and $PASSWORD as environment variables or is there a possibility that the plug in can pass these to the environment of the exec tasks?

saw303: Your PR will create a docker login command that containing the effective values for username and password. If you execute the command will log that values for username and password. This is a security issue since in a public CI environment everyone could then read your credentials for a Docker registry.

@iamdanfox suggest to move the credentials to environment variables. That way the effective credential values aren't logged by the console.

chris060986:
@iamdanfox ok now?

pkoenig10:
It would be nice if we could generalize this to support logging in to multiple private repositories.

chris060986:
@pkoenig10 
See also [Pull Request 219](https://github.com/palantir/gradle-docker/pull/219).

pkoenig10:
@chris060986 should we close this in favor of that PR?

chris060986:
> @chris060986 should we close this in favor of that PR?

yeah, depends on you, what is more interesting to merge into the develop branch...

recursivecodes:
Is there any way to do this today?  How can I push to a private repo?

chris060986:
@recursivecodes
Clone the repository, may change the plugin Id, build and publish the plugin to maven local. Then you can use it as described in the read me. That's also the way I'm using it at the moment.

recursivecodes:
> @recursivecodes
> Clone the repository, may change the plugin Id, build and publish the plugin to maven local. Then you can use it as described in the read me. That's also the way I'm using it at the moment.

Is there a reason this isn't being merged into the project?  Seems like it would be a common requirement?

saumilsdk:
Is this supported? As i am still not able to see any login commands available for this plugin.

dsinghvi:
Yup +1

Issue #195, opened 2018-08-23
Improvement: Allow dockerPush to target multiple repositories


BenedictRasmussen:
We have a use case where we need to create an image and push it to multiple repositories To accomplish this now, we create a single image, tag it multiple times, then push to all repositories. For example:

```
docker build -t repository01.com:2018.08.23 .
docker tag repository01.com:2018.08.23 repository02.com:2018.08.23
docker tag repository01.com:2018.08.23 repository03.com:2018.08.23

docker push repository01.com:2018.08.23
docker push repository02.com:2018.08.23
docker push repository03.com:2018.08.23
```

Currently `gradle-docker` can accept multiple tags when creating an image, but this is only applied to the actual docker tag after the colon. I am proposing allowing users to specify multiple names in addition to multiple tags. Each tag would then be applied to each name. When the user runs `gradle dockerPush`, it `clean`s / `prepare`s / `exec`s / as usual. During the tag step, the `name`s list is iterated and a tag applied to each name, then during the `push` step, every <name>:<tag> pair is pushed.

This is something I am willing to work on in my spare time if the maintainers think this would be a useful feature.

devkanro:
See this PR #205 , it can create multiple repositories image and push it

pkoenig10:
Fixed in #205

Issue #181, opened 2018-06-13
Pushing just the defined tags


pluttrell:
Using version `0.19.2` with the following script snippet, when we run the `dockerPush` task we're seeing the plugin push _all_ images, not just the version specified in `tags` and the `latest` as defined in `docker.tags`. Is there a way to get it to only push these two tags?

```groovy
docker.name = 'mycompany/myname'

docker {
    tags project.version, 'latest'
    /* ... other items omitted for brevity ... */
}

tasks.dockerPush.dependsOn { tasks.dockerTag }

dockerPush {
    dependsOn { tasks.dockerTag }
}
```

kerko:
@pluttrell 
Simple solution that does not need support from the Plugin Authors, add the following to your build script:
```
private static String ucfirst(String str) {
        StringBuffer sb = new StringBuffer(str);
        sb.replace(0, 1, str.substring(0, 1).toUpperCase());
        return sb.toString();
    }

task dockerPublish {
  for(tag in  docker.tags)
  dependsOn "dockerPush"+ucfirst(tag)
}
```
now running dockerPublish should only push the tags you mentioned

BenedictRasmussen:
Just want to second this issue. I used the following configuration:

```
docker {
    tags "bras-test1"
    ...
}
```

When I ran `gradle dockerPush` and checked my repository after, the repository contained `bras-test1`, `bras-test2`, `bras-test3`, `bras-test4`, and `bras-test`, most of which were test images I'd create at a previous time.

devkanro:
See this PR #205 , and use `dockerTagsPush` task to push taged image only, `dockerAllPush` task to push all

PR #205, opened 2018-10-17
Support tag task name for 'tags'

devkanro:
## Description
This PR make 'tags' more powerful, you can use new 'tag' DSL to give the tag a task name.
There are some samples:
```Groovy
docker {
    name 'fake-service-name'
    tag 'withTaskName', 'fake-service-name:2.0'
    tag 'newImageName', 'new-image:latest'
}
```
It will generate tag tasks with name `dockerTagWithTaskName` and `dockerTagNewImageName`, and push tasks with name `dockerPushWithTaskName` and `dockerPushNewImageName`.

With task name, we can build some diff name image of one image, push to diff repo host.

## Note
'tags' DSL can resolve the tag name auto, but 'tag' can not. It means you must specify the image name for 'tag' DSL.  

There are some sample:
**Tag with just new version:**
name: `old-name:test`
tags dsl: `tags 'latest'`
tag dsl: `tag 'latest', 'docker-name:latest'`
image `result name: docker-name:latest`

**Tag with new name:**
name: `old-name:test`
tags dsl: `tags 'newName@docker-name:latest'`
tag dsl: `tag 'newName', 'docker-name:latest'`
image `result name: docker-name:latest`

**Tag with new repo host:**
name: `old-name:test`
tag: `newName@host:port/repo/docker-name:latest`
tags: `newName', 'host:port/repo/docker-name:latest'`
image result name: `host:port/repo/docker-name:latest`

>**Note:**
>Notice that tag must have a task name if you want to replace the repo or image name for 'tags' DSL

## New tasks
**dockerTagsPush**
Pushes all tagged Docker images to configured Docker Hub.
```
dockerTagsPush= dockerPush<Tag1>+ dockerPush<Tag2> + ... + dockerPush<Tagn>
```

**dockerAllPush**
Pushes all tagged Docker images and named Docker image to configured Docker Hub.
```
dockerAllPush = dockerTagsPush + dockerPush
```

## Solved issues
#195 Allow dockerPush to target multiple repositories
Use `dockerPush<Tag>` to diff repositories
```Groovy
docker {
    name 'repository1.foo.com/bar:lastest'
    tag 'repository2', 'repository2.foo.com/bar:lastest'
}
```
Run this command:
```Shell
./gradle dockerPushRepository2
```

#181 Pushing just the defined tags
Use `dockerTagsPush` task can solve this issue
```Groovy
docker {
    name 'repository1.foo.com/bar:lastest'
    tag 'tag1', 'bar:v1'
    tag 'tag2', 'bar:v2'
}
```
Run this command:
```Shell
./gradle dockerTagsPush
```

#95 Make push task depend on push tasks for each tag if tags are present
Use `dockerTagsPush` and `dockerPush` can solve this issue/PR without change behavior of `dockerPush`
```Groovy
docker {
    name 'repository1.foo.com/bar:lastest'
    tag 'tag1', 'bar:v1'
    tag 'tag2', 'bar:v2'
}
```
Run this command:
```Shell
./gradle dockerPush
./gradle dockerTagsPush
```

#149 dockerPush doesn't generate (and thus push) all Tags
#94 Ability to easily push all tags for an image
#209 Docker tag should support full retagging

## Test sample
There are some sample which used in test:

name | tag | result
------------ | ------------- | -------------
v1 | latest | v1:latest
v1:1 | latest | v1:latest
host/v1 | latest | host/v1:latest
host/v1:1 | latest | host/v1:latest
host:port/v1 | latest | host:port/v1:latest
host:port/v1:1 | latest | host:port/v1:latest
v1 | name@latest | v1:latest
v1:1 | name@latest | v1:latest
host/v1 | name@latest | host/v1:latest
host/v1:1 | name@latest | host/v1:latest
host:port/v1 | name@latest | host:port/v1:latest
host:port/v1:1 | name@latest | host:port/v1:latest
v1 | name@v2:latest | v2:latest
v1:1 | name@v2:latest | v2:latest
host/v1 | name@v2:latest | v2:latest
host/v1:1 | name@v2:latest | v2:latest
host:port/v1 | name@v2:latest | v2:latest
host:port/v1:1 | name@v2:latest | v2:latest
v1 | name@host/v2 | host/v2
v1:1 | name@host/v2 | host/v2
host/v1 | name@host/v2 | host/v2
host/v1:1 | name@host/v2 | host/v2
host:port/v1 | name@host/v2 | host/v2
host:port/v1:1 | name@host/v2 | host/v2
v1 | name@host/v2:2 | host/v2:2
v1:1 | name@host/v2:2 | host/v2:2
host/v1 | name@host/v2:2 | host/v2:2
host/v1:1 | name@host/v2:2 | host/v2:2
host:port/v1 | name@host/v2:2 | host/v2:2
host:port/v1:1 | name@host/v2:2 | host/v2:2
v1 | name@host:port/v2:2 | host:port/v2:2
v1:1 | name@host:port/v2:2 | host:port/v2:2
host/v1 | name@host:port/v2:2 | host:port/v2:2
host/v1:1 | name@host:port/v2:2 | host:port/v2:2
host:port/v1 | name@host:port/v2:2 | host:port/v2:2
host:port/v1:1 | name@host:port/v2:2 | host:port/v2:2

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -32,7 +32,7 @@ automatically include outputs of task dependencies in the Docker build context.
 - `name` the name to use for this container, may include a tag
 - `tags` (deprecated) (optional) an argument list of tags to create; any tag in `name` will
   be stripped before applying a specific tag; defaults to the empty set
-- `tag` (optional) defined a tag with task name
+- `tag` (optional) a tag to create with a specified task name
 - `dockerfile` (optional) the dockerfile to use for building the image; defaults to
   `project.file('Dockerfile')` and must be a file object
 - `files` (optional) an argument list of files to be included in the Docker build context, evaluated per `Project#files`. For example, `files tasks.distTar.outputs` adds the TAR/TGZ file produced by the `distTar` tasks, and `files tasks.distTar.outputs, 'my-file.txt'` adds the archive in addition to file `my-file.txt` from the project root directory. The specified files are collected in a Gradle CopySpec which may be copied `into` the Docker build context directory. The underlying CopySpec may also be used to copy entire directories into the build context. The following example adds the aforementioned archive and text file to the CopySpec, uses the CopySpec to add all files `from` `src/myDir` into the CopySpec, then finally executes the copy `into` the docker build context directory `myDir`
@@ -82,7 +82,7 @@ Configuration specifying all parameters:
 docker {
     name 'hub.docker.com/username/my-app:version'
     tags 'latest' // deprecated, use 'tag'
-    tag 'privateRepo', 'my.repo.com/username/my-app:version'
+    tag 'myRegistry', 'my.registry.com/username/my-app:version'
     dockerfile file('Dockerfile')
     files tasks.distTar.outputs, 'file1.txt', 'file2.txt'
     buildArgs([BUILD_VERSION: 'version'])
@@ -259,7 +259,7 @@ Tasks
    * `dockerTag<tag>`: tag the docker image with `<tag>`
    * `dockerPush`: push the specified image to a docker repository
    * `dockerPush<tag>`: push the `<tag>` docker image to a docker repository
-   * `dockerTagsPush`: push all tagged Docker images to configured Docker Hub
+   * `dockerTagsPush`: push all tagged docker images to a docker repository
    * `dockerPrepare`: prepare to build a docker image by copying
      dependent task outputs, referenced files, and `dockerfile` into a temporary
      directory
--- a/readme.md
+++ b/readme.md
@@ -30,8 +30,9 @@ automatically include outputs of task dependencies in the Docker build context.
 
 **Docker Configuration Parameters**
 - `name` the name to use for this container, may include a tag
-- `tags` (optional) an argument list of tags to create; any tag in `name` will
+- `tags` (deprecated) (optional) an argument list of tags to create; any tag in `name` will
   be stripped before applying a specific tag; defaults to the empty set
+- `tag` (optional) defined a tag with task name
 - `dockerfile` (optional) the dockerfile to use for building the image; defaults to
   `project.file('Dockerfile')` and must be a file object
 - `files` (optional) an argument list of files to be included in the Docker build context, evaluated per `Project#files`. For example, `files tasks.distTar.outputs` adds the TAR/TGZ file produced by the `distTar` tasks, and `files tasks.distTar.outputs, 'my-file.txt'` adds the archive in addition to file `my-file.txt` from the project root directory. The specified files are collected in a Gradle CopySpec which may be copied `into` the Docker build context directory. The underlying CopySpec may also be used to copy entire directories into the build context. The following example adds the aforementioned archive and text file to the CopySpec, uses the CopySpec to add all files `from` `src/myDir` into the CopySpec, then finally executes the copy `into` the docker build context directory `myDir`
@@ -53,7 +54,7 @@ docker {
 To build a docker container, run the `docker` task. To push that container to a
 docker repository, run the `dockerPush` task.
 
-Tag and Push tasks for each tag will be generated for each provided `tags` entry.
+Tag and Push tasks for each tag will be generated for each provided `tag` and `tags` entry.
 
 **Examples**
 
@@ -80,7 +81,8 @@ Configuration specifying all parameters:
 ```gradle
 docker {
     name 'hub.docker.com/username/my-app:version'
-    tags 'latest'
+    tags 'latest' // deprecated, use 'tag'
+    tag 'privateRepo', 'my.repo.com/username/my-app:version'
     dockerfile file('Dockerfile')
     files tasks.distTar.outputs, 'file1.txt', 'file2.txt'
     buildArgs([BUILD_VERSION: 'version'])
@@ -257,6 +259,8 @@ Tasks
    * `dockerTag<tag>`: tag the docker image with `<tag>`
    * `dockerPush`: push the specified image to a docker repository
    * `dockerPush<tag>`: push the `<tag>` docker image to a docker repository
+   * `dockerTagsPush`: push all tagged Docker images to configured Docker Hub
+   * `dockerAllPush`: push all tagged Docker images and named Docker image to configured Docker Hub
    * `dockerPrepare`: prepare to build a docker image by copying
      dependent task outputs, referenced files, and `dockerfile` into a temporary
      directory
--- a/readme.md
+++ b/readme.md
@@ -260,7 +260,6 @@ Tasks
    * `dockerPush`: push the specified image to a docker repository
    * `dockerPush<tag>`: push the `<tag>` docker image to a docker repository
    * `dockerTagsPush`: push all tagged Docker images to configured Docker Hub
-   * `dockerAllPush`: push all tagged Docker images and named Docker image to configured Docker Hub
    * `dockerPrepare`: prepare to build a docker image by copying
      dependent task outputs, referenced files, and `dockerfile` into a temporary
      directory
--- a/src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
@@ -19,6 +19,7 @@ import com.google.common.base.Preconditions
 import com.google.common.base.Strings
 import com.google.common.collect.ImmutableMap
 import com.google.common.collect.ImmutableSet
+import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.file.CopySpec
@@ -32,7 +33,8 @@ class DockerExtension {
     private String dockerComposeTemplate = 'docker-compose.yml.template'
     private String dockerComposeFile = 'docker-compose.yml'
     private Set<Task> dependencies = ImmutableSet.of()
-    private Set<String> tags = ImmutableSet.of()
+    private Set<String> unresolvedTags = new HashSet<>()
+    private Map<String, String> tags = new HashMap<>()
     private Map<String, String> labels = ImmutableMap.of()
     private Map<String, String> buildArgs = ImmutableMap.of()
     private boolean pull = false
@@ -85,12 +87,25 @@ class DockerExtension {
         copySpec.from(files)
     }
 
-    public Set<String> getTags() {
-        return tags
+    public Set<String> getUnresolvedTags() {
+        return unresolvedTags
     }
 
+    @Deprecated
     public void tags(String... args) {
-        this.tags = ImmutableSet.copyOf(args)
+        this.unresolvedTags.addAll(args)
+    }
+
+    public Map<String, String> getTags() {
+        return ImmutableMap.copyOf(tags)
+    }
+
+    public void tag(String taskName, String tag) {
+        if (tags.containsKey(taskName)) {
+            throw new GradleException("Task name '${taskName}' of docker tag '${tag}' is existed.")
+        }
+
+        tags.put(taskName, tag)
     }
 
     public Map<String, String> getLabels() {
--- a/src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
@@ -23,6 +23,8 @@ import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.file.CopySpec
+import org.gradle.internal.logging.text.StyledTextOutput
+import org.gradle.internal.logging.text.StyledTextOutputFactory
 
 class DockerExtension {
     Project project
@@ -33,8 +35,8 @@ class DockerExtension {
     private String dockerComposeTemplate = 'docker-compose.yml.template'
     private String dockerComposeFile = 'docker-compose.yml'
     private Set<Task> dependencies = ImmutableSet.of()
-    private Set<String> unresolvedTags = new HashSet<>()
-    private Map<String, String> tags = new HashMap<>()
+    private Set<String> tags = ImmutableSet.of()
+    private Map<String, String> namedTags = new HashMap<>()
     private Map<String, String> labels = ImmutableMap.of()
     private Map<String, String> buildArgs = ImmutableMap.of()
     private boolean pull = false
@@ -87,25 +89,24 @@ class DockerExtension {
         copySpec.from(files)
     }
 
-    public Set<String> getUnresolvedTags() {
-        return unresolvedTags
+    public Set<String> getTags() {
+        return tags
     }
 
     @Deprecated
     public void tags(String... args) {
-        this.unresolvedTags.addAll(args)
+        this.tags = ImmutableSet.copyOf(args)
     }
 
-    public Map<String, String> getTags() {
-        return ImmutableMap.copyOf(tags)
+    public Map<String, String> getNamedTags() {
+        return ImmutableMap.copyOf(namedTags)
     }
 
     public void tag(String taskName, String tag) {
-        if (tags.containsKey(taskName)) {
-            throw new GradleException("Task name '${taskName}' of docker tag '${tag}' is existed.")
+        if (namedTags.putIfAbsent(taskName, tag) == null) {
+            StyledTextOutput o = project.services.get(StyledTextOutputFactory.class).create(DockerExtension)
+            o.withStyle(StyledTextOutput.Style.Error).println("WARNING: Task name '${taskName}' of docker tag '${tag}' is existed.")
         }
-
-        tags.put(taskName, tag)
     }
 
     public Map<String, String> getLabels() {
--- a/src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
@@ -103,9 +103,9 @@ class DockerExtension {
     }
 
     public void tag(String taskName, String tag) {
-        if (namedTags.putIfAbsent(taskName, tag) == null) {
+        if (namedTags.putIfAbsent(taskName, tag) != null) {
             StyledTextOutput o = project.services.get(StyledTextOutputFactory.class).create(DockerExtension)
-            o.withStyle(StyledTextOutput.Style.Error).println("WARNING: Task name '${taskName}' of docker tag '${tag}' is existed.")
+            o.withStyle(StyledTextOutput.Style.Error).println("WARNING: Task name '${taskName}' is existed.")
         }
     }
 
--- a/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
@@ -138,16 +138,16 @@ class PalantirDockerPlugin implements Plugin<Project> {
             }
 
             if (!ext.unresolvedTags.isEmpty()) {
-                ext.unresolvedTags.each { markedTagName ->
-                    String taskName = generateTagTaskName(markedTagName)
+                ext.unresolvedTags.each { unresolvedTagName ->
+                    String taskName = generateTagTaskName(unresolvedTagName)
 
                     if (tags.containsKey(taskName)) {
-                        throw new GradleException("Task name '${taskName}' of docker tag '${tagName}' is existed.")
+                        throw new GradleException("Task name '${taskName}' of docker tag '${unresolvedTagName}' is existed.")
                     }
 
                     tags[taskName] = [
-                            rawTagName          : markedTagName,
-                            resolveTagNameAction: { -> computeName(ext.name, markedTagName) }
+                            rawTagName          : unresolvedTagName,
+                            resolveTagNameAction: { -> computeName(ext.name, unresolvedTagName) }
                     ]
                 }
             }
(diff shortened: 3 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @higankanshi! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

devkanro:
Any update here? @pkoenig10 
Not just me, there are still many people who really need this feature.

devkanro:
@uschi2000 @pkoenig10 
Could you review this PR? I really need this feature.
if it still no reply for me, I have to publish my fork to plugin repositories.

uschi2000:
Hej @higankanshi , sorry this slipped through the cracks. I'll take a look now.

devkanro:
@uschi2000 Thanks for your time.

uschi2000:
Thinking about this change a bit, I wonder if we should take a harder break and rectify some of the config birth defects, in particular the asymmetry between image `name` and image tags. Should they really be the same thing, and thus naturally allow for multiple hosts etc?
Something like
```
docker {
  imageName repo:port/foo:1.0
  imageName otherRepo/foo:latest, myTaskName
  ..
}
```

My rationale is that while your proposed is clearly something we should do (in order to fix all those issues), the implementation makes the config language pretty complicated (as witnessed by constraints like `Notice that tag must have a task name if you want replace the repo or image name`).

Thoughts?

devkanro:
@uschi2000 
We can use the [tag](https://docs.docker.com/engine/reference/commandline/tag/#extended-description) command of docker to set repo host for docker, like this:
```shell
docker tag httpd fedora/httpd:version1.0
docker tag httpd:test fedora/httpd:version1.0.test
```

So, i think that tag can be include a repo host, new image name.
But use the @ for marking task name, I think it can be more better, like your example
```groovy
docker {
  tag 'repo:port/foo:1.0' // Invalid tag, because we can get a valid name for this tag
  tag 'foo:latest' // Maybe can make a dockerFooLatest task ?
  tag 'latest' // Maybe we should not support version only for defined a tag
  tag 'otherRepo/foo:latest', myTaskName // Good, a custom name
}
```

But if we do this, it would be a breaking change for this plugin...
We should seriously weigh for it.

Using @ for backward compatible
or
Using new dsl for better semantics

pkoenig10:
I agree with @uschi2000.

It's seems odd to me that `tag` in this plugin is interpreted differently than arguments to the `docker tag`. Namely, we try to merge the `name` and the `tag` fields to compute an argument for `docker tag`.

I think we still need the `name` field to provide a primary identifier for the created image that can be used by other tasks. But the `name` and `tag` fields would be interpreted identically. The only difference being that the `name` tag is applied at build time, while the `tag` tags are applied in a separate task.

Providing custom task names seems like a step in the right direction, since it seems hopeless for us to try and derive task names if given tags which may or may not contain a registry, repository, image name, and/or tag.

devkanro:
@pkoenig10 
Yes, you are right. 

How do you think about using @ for marking task name?
I am using this for backward compatible, you do not have to change your existing config for this change.

uschi2000:
@pkoenig10 if you want to do the DSL break, then we'd also have to think about excavating our internal repos to get them towards the new config. should be relatively simple, but just flagging.

pkoenig10:
Ideally I'd like to avoid a DSL break here. Using @ for marking task name seems like a reasonable approach. This should be safe since @ cannot be used in tag names.

It looks like the primary blocker to supporting the better DSL is the `tags` method that takes a variable number of tags. Maybe as part of this change we should also deprecate this method.

Eventually it would be good to push everyone to the better DSL. This probably will take a couple of steps:
1. Support using @ for marking task names and deprecated the varargs tags method
2. [break] Remove the varargs tags method, replace it with a two arg (tag, taskName) tags method

And then optionally, to remove custom syntax from this plugin:
3. [break] Remove support for using @ to mark task names. This break will fail loudly since @ cannot be used in tag names.

devkanro:
OK, got it, I will change my PR later

devkanro:
@uschi2000 @pkoenig10 
I have added the 'tag' DSL and marked the 'tags' DSL as deprecated.

```Groovy
docker {
  tag 'newTag', 'repo:port/foo:1.0' 
}
```
However, it doesn't resolve image name for the tag. If you want to retag the version, you must specify the image name too.
This limit due to the design of docker's tag command, it just makes them as same.

```Groovy
docker {
  tag 'latest', 'latest'  // Invalid tag, it must be "tag 'latest', 'name:latest' "
}
```

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -278,7 +285,7 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
         """.stripIndent()
 
         when:
-        BuildResult buildResult = with('dockerTag').build()
+        BuildResult buildResult = with('dockerTag', "--stacktrace").build()
```
pkoenig10: Don't think we need `--stacktrace` here

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
```diff
@@ -85,12 +87,25 @@ class DockerExtension {
         copySpec.from(files)
     }
 
-    public Set<String> getTags() {
-        return tags
+    public Set<String> getUnresolvedTags() {
```
pkoenig10: Renaming this method is an API break. Maybe rename the new method to something like `getNamedTags`?

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
```diff
@@ -85,12 +87,25 @@ class DockerExtension {
...
+        return unresolvedTags
     }
 
+    @Deprecated
     public void tags(String... args) {
-        this.tags = ImmutableSet.copyOf(args)
+        this.unresolvedTags.addAll(args)
```
pkoenig10: I don't think we need/want to change the behavior of this method. Specifically it should continue to overwrite any previously set tags.

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
```diff
@@ -85,12 +87,25 @@ class DockerExtension {
...
+    }
+
+    public Map<String, String> getTags() {
+        return ImmutableMap.copyOf(tags)
+    }
+
+    public void tag(String taskName, String tag) {
+        if (tags.containsKey(taskName)) {
```
pkoenig10: Consider using `putIfAbsent` instead

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -85,6 +85,17 @@ class PalantirDockerPlugin implements Plugin<Project> {
             dependsOn tag
         })
 
+        Task pushAllTags = project.tasks.create('dockerTagsPush', {
+            group = 'Docker'
+            description = 'Pushes all tagged Docker images to configured Docker Hub.'
+        })
+
+        Task pushAll = project.tasks.create('dockerAllPush', {
```
pkoenig10: I think these "push all" workflows are already supported by the `dockerPush` task.

As long as `name` does not include a tag, then `name` is the docker repository and `docker push myname` will push the image and all tags.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -176,29 +208,68 @@ class PalantirDockerPlugin implements Plugin<Project> {
...
+        if (firstAt > 0) {
+            // Get substring of task name
+            tagTaskName = name.substring(0, firstAt)
+        } else if (firstAt == 0) {
+            // Task name must not be empty
+            throw new GradleException("Task name of docker tag '${name}' must not be empty.\n" +
+                    "There are some sample:\n" +
+                    "firstVersion@1.0.0 (dockerTagFirstVersion)\n" +
+                    "newName@docker-name:latest (dockerTagNewName)\n" +
+                    "withRepo@myregistryhost/fedora/httpd:version1.0 (dockerTagWithRepo)")
```
pkoenig10: I'd like to avoid specific examples in the source code. I think it's sufficient here to just have a clear error message.

The README is probably a better place for these examples.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -119,32 +130,53 @@ class PalantirDockerPlugin implements Plugin<Project> {
                 logging.captureStandardError LogLevel.ERROR
             }
 
-            if (!ext.tags.isEmpty()) {
+            Map<String, Object> tags = ext.tags.collectEntries { taskName, tagName ->
+                [generateTagTaskName(taskName), [
+                        rawTagName          : tagName,
+                        resolveTagNameAction: { -> tagName }
```
pkoenig10: I would prefer to make this a defined immutable object, but that might have to wait until we convert this to Java

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -176,29 +208,68 @@ class PalantirDockerPlugin implements Plugin<Project> {
...
 
-    private static String ucfirst(String str) {
-        StringBuffer sb = new StringBuffer(str);
-        sb.replace(0, 1, str.substring(0, 1).toUpperCase());
-        return sb.toString();
+            return name.substring(0, endIndex) + ":" + tagValue
+        }
     }
 
+    @Deprecated
```
pkoenig10, line 240: Is this supposed to be deprecated?

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
pkoenig10: nit: rename `rawTagName` and `resolveTagNameAction` to `tagName` and `tagTask` for clarity

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -119,32 +130,53 @@ class PalantirDockerPlugin implements Plugin<Project> {
...
+                ext.unresolvedTags.each { unresolvedTagName ->
+                    String taskName = generateTagTaskName(unresolvedTagName)
+
+                    if (tags.containsKey(taskName)) {
+                        throw new GradleException("Task name '${taskName}' of docker tag '${unresolvedTagName}' is existed.")
+                    }
+
+                    tags[taskName] = [
+                            rawTagName          : unresolvedTagName,
+                            resolveTagNameAction: { -> computeName(ext.name, unresolvedTagName) }
```
pkoenig10: Why make this a closure? We're almost certainly going to evaluate it so I'd prefer to just eagerly evaluate it to make this simpler.

pkoenig10 (commented):
Design looks good, couple of implementation comments

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
devkanro: It is just for test, will remove it

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
devkanro: Ok, but I think we will delete the tags dsl finally.

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
devkanro: Ok, got it

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
devkanro: Got it, I am a groovy noob, so sometimes I am using groovy like java

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
devkanro: Got it.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
devkanro: Yes, `tags` DSL will be deprecated in future, we will no longer need this method.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
devkanro: Got it.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
devkanro: Due to this test 'does not throw if name is configured after evaluation phase'.
```Groovy
docker {
    tags 'latest', 'another', 'withTaskName@2.0', 'newImageName@${id}-new:latest'
    tag 'withTaskNameByTag', '${id}:new-latest'
}
afterEvaluate {
    docker.name = '${id}'
}
```

You can set the docker name in `afterEvaluate`, if we resolve the tag immediately, we could not get the name of the docker, and the name of docker is required for resolve tag, so it will throw an exception.
So, I create a closure for resolving it.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
devkanro: To some extent this has caused some confusion and confusion. `dockerPush` can push a part of tags or not push. We need some clarity task for push tags.

I think the best is `dockerPush` push the main docker, will not push the tags.
`dockerTagsPush` will just push the tags, will not push the main docker.
`dockerAllPush` will push both of the main docker and the tags.

But if we change the behavior of `dockerPush`, it will be another breaking change. So I kept its original action. Just add new tasks for clarity.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
devkanro: I can't make this map as immutable, because unresolved tag will add to this map later.

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
pkoenig10: We probably don't need to mark private methods as deprecated, but this is fine

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
pkoenig10: This seems reasonable. It is a bit weird since `dockerPush` might behave the same as `dockerAllPush` if the `name` field does not include a tag. But we can't really avoid this with the current design.

This really just highlights how misleading the `name` field is. Ultimately, I think it would be preferable to treat the `name` field like any other `tag` and apply all the tags when building the image (see https://github.com/moby/moby/pull/15780).

A couple concrete comments:
1. Adding these tasks is unrelated to the improve DSL, can we move this change to a separate PR?
2. What's the use case for `dockerTagsPush`? It seems a bit confusing to me. I think we probably just need something like `dockerPush<tag>` and `dockerPushAll`.

devkanro: First, we can't use `dockerPushAll`, due to it has a conflict with `dockerPush<tag>` when you have a tagging task named 'all'. But maybe we can some names as reserved like 'all', 'tags'?

In fact, I am using `dockerTagsPush` now.
I have two repositories to push, I named the main dock as `imageName:version`.
And I have two tags, `repostiory1/imageName:version` and `repostiory2/imageName:version`, so I named those tag as `repostiory1` and `repostiory2`.
I push them in our CI system by `dockerTagsPush`, and in some case like debug for docker, I push one of them manually by `dockerPushRepostiory1`.

I will never push the main docker which defined by the name; it could be some confusion if I set the repository to name and using `dockerPush` to push one.
Why I push repository1 using `dockerPushRepostiory1`, but repository2 is `dockerPush`? Should it be `dockerPushRepostiory2`?

So I create the `dockerTagsPush` and `dockerAllPush` task.

pkoenig10: I would be in favor of adding `dockerTagsPush` but not `dockerAllPush`. This will allow us to move closer toward making `name` and `tags` behave the same (since under the hood they are both just tags).

So the image tags can be set and pushed by either:
- Using the `name` field and the `dockerPush` task
- Using the `tag` field and either `dockerPushMyTaskName` or `dockerTagsPush`

devkanro: OK, you are right, I have removed the `dockerAllPush` task.

readme.md
```diff
@@ -30,8 +30,9 @@ automatically include outputs of task dependencies in the Docker build context.
 
 **Docker Configuration Parameters**
 - `name` the name to use for this container, may include a tag
-- `tags` (optional) an argument list of tags to create; any tag in `name` will
+- `tags` (deprecated) (optional) an argument list of tags to create; any tag in `name` will
   be stripped before applying a specific tag; defaults to the empty set
+- `tag` (optional) defined a tag with task name
```
pkoenig10: nit: "a tag to create with a specified task name"

readme.md
```diff
@@ -80,7 +81,8 @@ Configuration specifying all parameters:
 ```gradle
 docker {
     name 'hub.docker.com/username/my-app:version'
-    tags 'latest'
+    tags 'latest' // deprecated, use 'tag'
+    tag 'privateRepo', 'my.repo.com/username/my-app:version'
```
pkoenig10: nit: replace "repo" with "registry":
```
tag 'myRegistry`, `my.registry.com/username/my-app:version`
```

readme.md
```diff
@@ -257,6 +259,7 @@ Tasks
    * `dockerTag<tag>`: tag the docker image with `<tag>`
    * `dockerPush`: push the specified image to a docker repository
    * `dockerPush<tag>`: push the `<tag>` docker image to a docker repository
+   * `dockerTagsPush`: push all tagged Docker images to configured Docker Hub
```
pkoenig10: nit: "push all tagged docker images to a docker repository"

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -119,32 +126,54 @@ class PalantirDockerPlugin implements Plugin<Project> {
...
+                ]]
+            }
+
             if (!ext.tags.isEmpty()) {
+                ext.tags.each { unresolvedTagName ->
+                    String taskName = generateTagTaskName(unresolvedTagName)
+
+                    if (tags.containsKey(taskName)) {
+                        StyledTextOutput o = project.services.get(StyledTextOutputFactory.class).create(DockerExtension)
+                        o.withStyle(StyledTextOutput.Style.Error).println("WARNING: Task name '${taskName}' is existed.")
```
pkoenig10: Let's just throw an `IllegalArgumentException` in this case

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -119,32 +126,54 @@ class PalantirDockerPlugin implements Plugin<Project> {
...
+                                tagTask: { -> computeName(ext.name, unresolvedTagName) }
+                        ]
+                    }
+                }
+            }
 
-                ext.tags.each { tagName ->
-                    String taskTagName = ucfirst(tagName)
-                    Exec subTask = project.tasks.create('dockerTag' + taskTagName, Exec, {
+            if (!tags.isEmpty()) {
```
pkoenig10: I don't think we need this if statement

pkoenig10 (commented):
LGTM, couple of minor comments

devkanro:
Fixed

src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -119,32 +126,51 @@ class PalantirDockerPlugin implements Plugin<Project> {
...
-                    project.tasks.create('dockerPush' + taskTagName, Exec, {
-                        group = 'Docker'
-                        description = "Pushes the Docker image with tag '${tagName}' to configured Docker Hub"
-                        workingDir dockerDir
-                        commandLine 'docker', 'push', "${ -> computeName(ext.name, tagName)}"
-                        dependsOn subTask
-                    })
+                    if (tags.containsKey(taskName)) {
+                        throw new IllegalArgumentException("Task name '${taskName}' is existed.")
+                    }else{
```
pkoenig10: nit: remove the else clause

pkoenig10 (approved):

pkoenig10:
@higankanshi can you merge/rebase this PR to pick up #220 so the checks pass?

PR #217, opened 2018-11-18
print docker output to console with default LogLevel.LIFECYCLE (fixes #216)

### LINKED ISSUE #216: Change default log level of docker output
Related to #127 

```bash
$ gradle docker
```

No output

---

```bash
$ gradle docker --info
```

Prints the output from `docker build` but also shows a lot of other unrelated output.

---

What do you think about changing the default stdout capturing to [LogLevel.LIFECYCLE](https://docs.gradle.org/current/userguide/logging.html#sec:choosing_a_log_level)?

When I run `docker build` on my shell, I see the progress output of the docker build steps by default. When I run `docker gradle`, I don't see it. I think it's a good idea to give the user feedback about the progress so that the docker steps are printed:

```
Sending build context to Docker daemon  25.73MB
Step 1/9 : FROM java:openjdk-8-jdk
(...and so on...)
```

https://github.com/palantir/gradle-docker/blob/develop/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy#L118

dherges:
#216

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
@@ -115,7 +115,7 @@ class PalantirDockerPlugin implements Plugin<Project> {
                 workingDir dockerDir
                 commandLine buildCommandLine(ext)
                 dependsOn ext.getDependencies()
-                logging.captureStandardOutput LogLevel.INFO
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @dherges! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #219, opened 2018-12-18
Allow login to multi docker repositories

chris060986:
Usage of multiple private docker repos to push or pull images.
Username and Password have to be set as environment variables.


uschi2000 (changes requested):
Yo Chris. I'd be fine with adding this feature, but I think it needs a few tweaks:

- I don't understand why we need to support logging into multiple repositories if the push is only ever to one?
- I'd prefer to pass in the repo config in the top-level `docker` config; looks like all we need there is username and password, so maybe something like `docker { credentials { user 'foo'\npassword 'bar' } }` would work?
- Let's avoid environment variables
- The whole thing needs test coverage
- The push tasks should depend on the `login` task iff creds are specified

chris060986:
> Let's avoid environment variables

In PR [198](https://github.com/palantir/gradle-docker/pull/198 ) your college suggested to use env variables instead of defining it in the gradle file

> I don't understand why we need to support logging into multiple repositories if the push is only ever to one?

I pull images from private repos but push into another private repo. Thats why I allow login to more than one repository

>The push tasks should depend on the login task iff creds are specified

Done!

> I'd prefer to pass in the repo config in the top-level docker config; looks like all we need there is username and password, so maybe something like docker { credentials { user 'foo'\npassword 'bar' } } would work?

In my opinion it's like the maven repos you can specify for a java build. Thats why its separated here. If it's realy important I may invest some more time.

I build the plugin on my own and use it with this feature.

uschi2000:
Chris, I was not aware of the other PR, sorry for the conflicting opinions. Let me discuss internally and get back to you with the suggested approach.

PR #231, opened 2019-03-12
Enable stalebot

pkoenig10:
## Before this PR
Github issues could accumulate, when the owner has lost interest or the problem has actually been fixed.

## After this PR
stalebot will automatically comment on any issues that haven't been touched in 60 days, and auto-close them if nobody cares anymore.

Matches the config we already have running on other repos.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,14 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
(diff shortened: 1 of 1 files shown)


iamdanfox:
@pkoenig10 we actually got quite a lot of pushback when we enabled this on conjure-java-runtime and baseline, and have since reverted it.

We found that stalebot only really worked if we _always_ prioritised fixing legitimate issues within the timeout, but occasionally we found that there are just higher priority internal things to work on, so legitimate issues actually should remain open.

I'm now trying it out for _PRs only_ in a few repos, and haven't had too much pushback yet.

Issue #168, opened 2018-02-20
copySpec and files instructions don't copy required files


kikulikov:
I have a project where I have a few folders and files which I need to include in the Docker build context. I've followed the tutorial from [gradle-docker](https://github.com/palantir/gradle-docker) but it doesn't seem to work for me. Only the last `copySpec` instruction is getting applied and no files are copied to the context. Please, see below the example of `build.gradle` configuration. Am I missing something? Is there any other way to do it?

```gradle
plugins {
	id "com.palantir.docker" version "0.19.2"
}

docker {
	name 'sample/demo-gradle-palantir-docker:version'
	
	files 'gradlew', 'build.gradle', 'settings.gradle'
	copySpec.from("src").into("src")
	copySpec.from("gradle").into("gradle")
}
```

A content of `build/docker/`

```sh
ls -A1 build/docker/
Dockerfile
src
```

A demo project with all `gradle` and `docker` files https://github.com/cyrillk/demo-gradle-palantir-docker

 Thank you.

pkoenig10:
A way to get the desired behavior is the following:
```
docker {
    name 'sample/demo-gradle-palantir-docker:version'

    files 'gradlew', 'build.gradle', 'settings.gradle'
    copySpec.with {
        from('.') {
            include 'src/**'
            into 'src'
        }
        from('.') {
            include 'gradle/**'
            into 'gradle'
        }
    }
}
```

The current implementation makes it awkward to copy directories. This is something that could probably be improved.

kikulikov:
Thanks! It seems to work ok.

wussow-everi:
This works great. I think it would be helpful if this information was present on the main page. I had to do some digging to find the answer to this.

sheikhmohamadarsalan:
managed to fix the issue of gradle, just want to share knowledge

In file build.gradle, I updated the line "files" to add oneagent.zip file to docker build context
docker {
    name "473076588230.dkr.ecr.ap-southeast-1.amazonaws.com/cbop/merchant-service:${version}"
    files tasks.bootJar.outputs, "build/libs/oneagent.zip"
    copySpec.into("build/libs")
    dockerfile file('Dockerfile')
    dependsOn tasks.bootJar
}

In buildspec.yml, I download file from S3 to build/libs
      - mkdir -p build/libs
      - aws s3 cp s3://op-pipeline/package/dynatrace/paasInstaller_Unix_all.zip build/libs/oneagent.zip
[...]

PR #238, opened 2019-04-10
Clarify doc on copySpec

AnEmortalKid:
update copy spec example with something that works

## Before this PR
Could not figure out how to use `copySpec`

## After this PR
A clear working example of copy spec is in the doc

Relevant to #168

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -35,18 +35,22 @@ automatically include outputs of task dependencies in the Docker build context.
 - `tag` (optional) a tag to create with a specified task name
 - `dockerfile` (optional) the dockerfile to use for building the image; defaults to
   `project.file('Dockerfile')` and must be a file object
-- `files` (optional) an argument list of files to be included in the Docker build context, evaluated per `Project#files`. For example, `files tasks.distTar.outputs` adds the TAR/TGZ file produced by the `distTar` tasks, and `files tasks.distTar.outputs, 'my-file.txt'` adds the archive in addition to file `my-file.txt` from the project root directory. The specified files are collected in a Gradle CopySpec which may be copied `into` the Docker build context directory. The underlying CopySpec may also be used to copy entire directories into the build context. The following example adds the aforementioned archive and text file to the CopySpec, uses the CopySpec to add all files `from` `src/myDir` into the CopySpec, then finally executes the copy `into` the docker build context directory `myDir`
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @AnEmortalKid! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

pkoenig10:
@AnEmortalKid can you elaborate on or provide an example for why the existing copy spec example didn't work for you? It seems to work as expected for me.

AnEmortalKid:
Ah, it might be not doing what I thought it would.

So, my desired state:

```
build/
  docker/
    'myfile.txt'
    data/
      'super-secret-password.txt'
```

Using:
```
  files 'myfile.txt'
  copySpec.from("secretData/").into("data")
```

I actually get:
```
build/
  docker/
    data/
      'myfile.txt'
      'super-secret-password.txt'
```

Maybe I misread the: 

> The specified files are collected in a Gradle CopySpec which may be copied into the Docker build context directory. The underlying CopySpec may also be used to copy entire directories into the build context. The following example adds the aforementioned archive and text file to the CopySpec, uses the CopySpec to add all files from src/myDir into the CopySpec, then finally executes the copy into the docker build context directory myDir

Particularly:

> The following example adds the aforementioned archive and text file to the CopySpec, uses the CopySpec to add all files from src/myDir into the CopySpec, then finally executes the copy into the docker build context directory myDir

Maybe if we clarify with the example output with regards to the `build/docker` that would help others? It didn't seem super obvious to me everything would end under `myDir`.

pkoenig10:
Yeah I think we can clarify the last sentence a bit.

> ... then finally executes the copy into the directory myDir in docker build context.

AnEmortalKid:
@pkoenig10 lmk if the output example is 2 much, and i can remove it. Hopefully the doc is more clear!

pkoenig10 (approved):
LGTM, thanks @AnEmortalKid!

PerMildner:
The "The final structure will be" in the updated documentation seems to document the current behavior but it does not answer the original question, i.e. how to get the non-flat result:

```
build/
  docker/
    'myfile.txt'
    data/
      'super-secret-password.txt'
```
It is unexpected, and not useful, that the `files` argument ends up in the `into`-folder used in a subsequent `copySpec...into(...)` call. Also, using `copySpec` with more than one `from` where the `from` has a closure with `exclude` seems to apply the `exclude` to the entire copy.

So, perhaps there is really some underlying bug in how copy specs are combined in the final copy. 

In either case the documentation would be more useful if it contained an example corresponding to the more complex structure above. It looks like https://github.com/palantir/gradle-docker/issues/168#issuecomment-375947722 may be relevant.

PR #239, opened 2019-04-10
[improvement] docker configuration includes productDependencies

iamdanfox:
## Before this PR

Internally, many repos use this plugin in conjunction with gradle-sls-packaging.  When trying to create a docker-compose.yml for ETE testing, users usually want to use versions of their _product dependencies_.

From [sls-packaging 3.7.1](https://github.com/palantir/sls-packaging/releases/tag/3.7.1) onwards, there's now a convenient way of getting access to this information.

## After this PR

Repos that use `com.palantir.docker-compose` at the root and `com.palantir.sls-java-service-distribution` in a subproject will now be able to magically refer to their product dependencies when templating their docker-compose.yml files:

```yml
 my-product:
    container_name: my-product.palantir.pt
    hostname: my-product
    domainname: palantir.pt
    image: 'docker-registry/foo/my-product:{{com.palantir.foo:my-product}}'
    ports:
      - "8901:8901"
      - "8902:8902"
```

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/DockerComposePlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerComposePlugin.groovy
@@ -41,6 +41,9 @@ class DockerComposePlugin implements Plugin<Project> {
                 )))
             })
         })
+        project.getPlugins().withId("com.palantir.product-dependency-introspection", {
+            dockerConfiguration.extendsFrom(project.configurations.getByName('productDependencies'))
+        })
 
         project.tasks.create('generateDockerCompose', GenerateDockerCompose, {
             it.ext = ext
--- a/src/main/groovy/com/palantir/gradle/docker/DockerComposePlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerComposePlugin.groovy
@@ -15,6 +15,7 @@
  */
 package com.palantir.gradle.docker
 
+import com.google.common.collect.ImmutableMap
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.artifacts.Configuration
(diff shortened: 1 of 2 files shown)


dansanduleac:
👍

src/main/groovy/com/palantir/gradle/docker/DockerComposePlugin.groovy
```diff
@@ -26,6 +27,24 @@ class DockerComposePlugin implements Plugin<Project> {
...
         Configuration dockerConfiguration = project.configurations.maybeCreate('docker')
 
+        // sls-packaging adds a 'productDependencies' configuration, which contains the inferred lower bounds of products
+        // you depend on.  We wire it up automatically, so all users don't need to add:
+        //
+        //    dependencies {
+        //        docker project(path: ':foo', configuration: 'productDependencies')
+        //    }
+        project.subprojects({ Project subproject ->
+            subproject.getPlugins().withId("com.palantir.product-dependency-introspection", {
```
iamdanfox, line 37: https://github.com/palantir/sls-packaging/blob/develop/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/ProductDependencyIntrospectionPlugin.java

qinfchen:
Does this mean product dependencies must be specified now?
```
* What went wrong:
Execution failed for task ':generateDockerCompose'.
> Could not resolve all dependencies for configuration ':docker'.
   > Could not resolve project :dummy-server.
     Required by:
         project :
      > Project : declares a dependency from configuration 'docker' to configuration 'productDependencies' which is not declared in the descriptor for project :.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

```

iamdanfox:
No, that looks like a bug.  Can you link the actual repro to the internal slack channel so we can poke around?

dansanduleac:
For future reference, this was due to the root project (where we defined the `docker` configuration being resolved) having the same `group:name` as a subproject. This is an artifact of how gradle dependency resolution works. 
The dependency `project : conf 'docker' -> project :subproject conf productDependencies` got resolved by gradle to `project : conf productDependencies` because `:` and `:subproject` had the same group and name, and because it had already seen and cached the group:name of `:` (the root project) since resolution started from there.

PR #241, opened 2019-04-17
Convert two classes to Java

iamdanfox:
## Before this PR

Currently, I am forced to use nasty afterEvaluate code to workaround the fact that this repo doesn't use providers.  As a pre-requisite to a provider refactor, I'd like to convert to Java so that the API surface of this plugin is a bit clearer (fewer magic methods).

The entire plugin is written in Groovy, but Gradle's [official advice](https://docs.gradle.org/current/userguide/custom_plugins.html) is:

> In general, a plugin implemented using Java or Kotlin, which are statically typed, will perform better than the same plugin implemented using Groovy.

## After this PR

Two classes have been converted to Java. This should provided marginally better performance, but more importantly it makes it easier to eyeball the API surface.


pkoenig10 (approved):

pkoenig10:
@iamdanfox do you still want this?

iamdanfox:
Nah don't worry about it

PR #243, opened 2019-04-18
[fix] Don't shadow the standard gradle 'ext' field on tasks.

GrahamDennis:
[fix] All Gradle Tasks (and many other things) have a special 'ext' extension,
which is of type ExtraPropertiesExtension
(see https://docs.gradle.org/current/javadoc/org/gradle/api/plugins/ExtensionAware.html).
Previously the DockerComposeExtension was shadowing this as it was set as a field
with name 'ext' on the gradle tasks created by this plugin.

## Before this PR

It was not possible to set custom properties on the `generateDockerCompose` task by the usual `generateDockerCompose.ext.foo = "bar"` method supported by all other Gradle tasks because the `GenerateDockerCompose` task had a field `ext` that contained the `DockerComposeExtension`.

## After this PR

This PR removes the `ext` field on `GenerateDockerCompose` and `DockerComposeUp` and gets the `DockerComposeExtension` directly from the project.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/DockerComposePlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerComposePlugin.groovy
@@ -23,8 +23,7 @@ import org.gradle.api.artifacts.Configuration
 class DockerComposePlugin implements Plugin<Project> {
     @Override
     void apply(Project project) {
-        DockerComposeExtension ext =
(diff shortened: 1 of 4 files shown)


pkoenig10 (approved):
Thanks @GrahamDennis!

GrahamDennis:
@iamdanfox: Can you approve this?

iamdanfox (approved):

PR #244, opened 2019-04-25
[improvement] Add support for --init as init property to dockerRun extension

christopherbryanboyd:
## Before this PR
The dockerRun plugin does a lot of what I need, but it does not have support for the --init flag.
The --init flag is necessary for processes to properly be terminated and reaped, as per the docker run man page:
```
--init
          Run an init inside the container that forwards signals and reaps processes
```

## After this PR
Now the dockerRun extension has a new boolean property `init`, default to false, but when set to true, `--init` is passed into the resulting `docker run` command.

A test has been included to verify the functionality.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -241,6 +241,7 @@ dockerRun {
     daemonize true
     env 'MYVAR1': 'MYVALUE1', 'MYVAR2': 'MYVALUE2'
     command 'sleep', '100'
+    init true
 }
 ```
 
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @christopherbryanboyd! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #245, opened 2019-04-25
[improvement] Add support for workdir (-w, --workdir)

christopherbryanboyd:
## Before this PR
There is currently no way to set the -w, --workdir from the dockerRun plugin.

Per the docker run man page:
```
Working directory inside the container
```

## After this PR
Now the working directory can be set with the syntax:
```
dockerRun {
     workdir '/dir'
}
```
Which will be where the dockerRun.command is executed.

This PR also includes a test for the functionality.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -241,6 +241,7 @@ dockerRun {
     daemonize true
     env 'MYVAR1': 'MYVALUE1', 'MYVAR2': 'MYVALUE2'
     command 'sleep', '100'
+    workdir '/containervolume'
 }
 ```
 
@@ -258,6 +259,7 @@ dockerRun {
 - `clean` (optional) a boolean argument which adds `--rm` to the `docker run`
   command to ensure that containers are cleaned up after running; defaults to `false`
 - `command` the command to run.
+- `workdir` the working directory inside the container; defaults to '/'.
 
 Tasks
 -----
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @christopherbryanboyd! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

akelix:
Very useful feature

christopherbryanboyd:
How strange, I'm not sure how this got closed!

dansanduleac:
The workdir can already be set inside the Dockerfile, and we control the dockerfile. Setting it there also means that other users pulling this image remotely get the right working dir, rather than having to manually override it.

Is there a case where this is not what you want?

christopherbryanboyd:
@dansanduleac There was a case where we needed this feature, but it's been so long that I don't remember why, and now we are just invoking docker manually from gradle through Exec tasks rather than using the plugin.

I think it had to do with requiring a different workdir for different invocations of the same Docker container.

dansanduleac:
Alright, fair enough. Thank you for explaining!

It sounds a bit esoteric compared to our usage, so I'm going to close this for now, but if you have a more thorough explanation of the motivation / what workflows this would enable, we can chat more.

PR #260, opened 2019-08-30
[fix]: Windows can test plugin

blacksun777:
Windows line ending \r\n caused regex matching problems. Also Windows \ file separator not compatible with docker for win.

## Before this PR

Unable to contribute from windows based machines as tests fail.

## After this PR

\R does for Java rgex matchers what \n does for Strings
However this is a Java 8 feature. We should bump it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,7 @@ if (System.env.CIRCLE_TEST_REPORTS) {
 group 'com.palantir.gradle.docker'
 version System.env.CIRCLE_TAG ?: gitVersion()
 
-sourceCompatibility = 1.7
+sourceCompatibility = 1.8
 
 task sourceJar(type: Jar) {
     from sourceSets.main.allSource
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @blacksun777! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

CRogers:
Thanks for the contribution! Looks like the source compatibility was explicitly set to Java 7 in #137, but since over 2 years has elapsed and we use Java 8 for pretty much all of our other gradle plugins it's probably OK to bump, provided there is a release note.

If we explicitly want to support Windows in the future it probably makes sense for us to set up a windows build on CircleCI to make sure it doesn't regress.

blacksun777:
Thank you for the feedback. If Java 7 is needed we can also do \r?\n matching, since it is not that important of a feature. (Should I do that? What are the Timelines for Java 11?)

Windows is only implicitly supported (Java), its not a real feature for Windows(tests only). There is already code in the tests to behave differently if it is run on Linux/CI. 
The test that uses `isWindows `is actually never run on CI. Therefore a Windows one would not be checked anyway. 

Where can i find the release notes/ change log?

When do the checks end? Policy is pending for a day and docs as well. Do I need to do something?

blacksun777:
dead repo or a nice way to say no?🤷‍♀️😉

PR #301, opened 2020-01-16
Upgradle

ferozco:
## Before this PR
This repo was pretty crufty with pretty much every aspect of its build out of date

## After this PR
No behaviour or code changes to the main source set. Only upgrading build tooling and fixing up some tests to get things to work with latest gradle.

Unfortunately I wasn't able to get this to use the circle template since basically all the tests require being on the machine executor

## Possible downsides?
Build tooling changes may break publishing or have impacted dependencies in some negative way, but probably unlikely


build.gradle
```diff
@@ -2,87 +2,50 @@ buildscript {
...
     }
-}
-buildscript {
-    repositories {
-        jcenter()
-        maven { url 'https://plugins.gradle.org/m2/' }
-        maven { url 'https://palantir.bintray.com/releases' }    }
     dependencies {
+        classpath 'com.netflix.nebula:gradle-info-plugin:5.2.0'
+        classpath 'com.palantir.javaformat:gradle-palantir-java-format:0.3.9'
```
dansanduleac, line 9: ❤️

circle.yml
```diff
@@ -1,30 +0,0 @@
...
-        shell: /bin/bash
-        command: ./gradlew --parallel -i build
-      - type: test-results-store
-        path: build/test-results
-      - type: deploy
-        shell: /bin/bash
-        command: ./gradlew -i bintrayUpload
-      - type: deploy
-        shell: /bin/bash
-        command: ./gradlew -i -Dgradle.publish.key=$GRADLE_KEY -Dgradle.publish.secret=$GRADLE_SECRET publishPlugins
```
dansanduleac, line 20: we need to add some code in build.gradle to set these:

https://github.com/palantir/sls-packaging/blob/2c285e6525f62632525c8f395e7240ea6cb0363c/gradle-sls-packaging/build.gradle#L83-L86

src/main/groovy/com/palantir/gradle/docker/DockerComponent.java
```diff
@@ -13,26 +13,29 @@
...
 import org.gradle.api.internal.attributes.ImmutableAttributesFactory;
 import org.gradle.api.internal.component.SoftwareComponentInternal;
 import org.gradle.api.internal.component.UsageContext;
 import org.gradle.api.model.ObjectFactory;
 
-import java.util.Collections;
-import java.util.LinkedHashSet;
-import java.util.Set;
-
 public class DockerComponent implements SoftwareComponentInternal {
```
dansanduleac, line 36: why do people do this :(

dansanduleac: Something for later: When implementing interfaces where methods might get added later, I recommend implementing them via a proxy. That way when they add something new, at least you *might* have a chance of not failing (if the new methods are never called).

build.gradle
```diff
@@ -106,7 +69,5 @@ pluginBundle {
     }
 }
 
-publishPlugins.onlyIf {
-    project.version ==~ /\d+\.\d+\.\d+/
-}
+publishPlugins.onlyIf { versionDetails().isCleanTag }
```
dansanduleac, line 72:
```suggestion
publishPlugins.onlyIf { versionDetails().isCleanTag }

// Configure the publishPlugins task
tasks.publish.dependsOn publishPlugins
project.ext.'gradle.publish.key' = System.env["GRADLE_KEY"]
project.ext.'gradle.publish.secret' = System.env["GRADLE_SECRET"]
```

dansanduleac:
👍

src/test/groovy/com/palantir/gradle/docker/PalantirDockerPluginTests.groovy
```diff
@@ -141,6 +142,10 @@ class PalantirDockerPluginTests extends AbstractPluginTest {
         execCond("docker rmi -f ${id}")
     }
 
+    // Gradle explicitly disallows the test case, fails with the following:
+    //Could not determine the dependencies of task ':publishDockerPublicationPublicationToMavenLocal'.
+    //> Publishing is not able to resolve a dependency on a project with multiple publications that have different coordinates.
+    @Ignore
```
iamdanfox, line 148: if we think this test is bogus maybe it should just get deleted instead? otherwise I don't know under what criteria is this going to get un-ignored?

iamdanfox (approved):
Nice cleanup!

PR #338, opened 2020-02-27
docker-run plugin - save container logs

wrekah:
## Before this PR
As I user, I would like to save container logs with palantir docker-run plugin. Right now functionality is missing in plugin and needs to be performed outside of it in dirty way.

## After this PR
* Adds new task in docker-run plugin called "dockerSaveContainerLog" that's executing "docker logs" and saving it to project build directory
* Unit test coverage for the new functionality

## Possible downsides?
* Tested this task with containers with log output up to 100M, possibly G+ logs should be tested 
* Executing task outside of container life cycle (i.e before it is started, after it is removed) fails, this is expected. Possibly some sanity check could be introduced to prevent such behaviour.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/DockerRunPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerRunPlugin.groovy
@@ -56,6 +56,11 @@ class DockerRunPlugin implements Plugin<Project> {
             description = 'Checks the network configuration of the container'
         })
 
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @tpiskorski! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #401, opened 2020-09-17
Improving null pointer exception with templateTokens

ben-pearson:
## Before this PR
A NullPointerException with no description/message/indication of what was wrong was thrown when supplying templateTokens with a valid key and a NULL value.

The 'docker-compose stop successfully stops docker container' test would fail if you're running any docker containers locally when running the test.

## After this PR
==COMMIT_MSG==
A more useful and friendly message is displayed when a null value is supplied in templateTokens.

The 'docker-compose stop successfully stops docker container' now just checks the container it brings up is also taken down - ignoring any other docker containers you may be running.
==COMMIT_MSG==

## Possible downsides?
None

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/DockerComposeExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerComposeExtension.groovy
@@ -15,6 +15,7 @@
  */
 package com.palantir.gradle.docker
 
+import static com.google.common.base.Preconditions.checkNotNull
 import com.google.common.collect.Maps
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @ben-pearson! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #513, opened 2021-12-13
Use project version as default tag

fawind:
## Before this PR
With https://github.com/palantir/gradle-docker/pull/506, we added a regression that not all docker images would be pushed if no tags were specified explicitly.

This is because the tasks `dockerTagsPush` and `dockerPush` are just marker tasks that run the respective tag and push tasks for all tags. However, when no tags are specified, no sub-tasks will be created and run as the below closure is skipped:

https://github.com/palantir/gradle-docker/blob/db4cfe13b99f703666247c0e17268318c58c58f9/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy#L148

## After this PR
To work-around this, we add the project version as a default tag, if not other tags are specified. This will per-default tag and push an image with the current project version.

==COMMIT_MSG==
Use project version as default tag
==COMMIT_MSG==

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
@@ -35,7 +35,7 @@ class DockerExtension {
     private String dockerComposeTemplate = 'docker-compose.yml.template'
     private String dockerComposeFile = 'docker-compose.yml'
     private Set<Task> dependencies = ImmutableSet.of()
-    private Set<String> tags = ImmutableSet.of();
+    private Set<String> tags = ImmutableSet.of()
     private Map<String, String> namedTags = new HashMap<>()
     private Map<String, String> labels = ImmutableMap.of()
     private Map<String, String> buildArgs = ImmutableMap.of()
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
```diff
@@ -90,12 +93,12 @@ class DockerExtension {
...
 
     public Set<String> getTags() {
-        return tags
+        return tags.get()
     }
 
     @Deprecated
     public void tags(String... args) {
-        this.tags = ImmutableSet.copyOf(args)
+        this.tags.set(ImmutableSet.copyOf(args))
```
CRogers: Hmm, from memory wasn't the way this used to work that people would put

```gradle
docker {
    tags 'latest'
}
```

but it would still make the version tag? I wonder if we just want to make `getTags()` return `tags.get() + [version]`?

fawind: Updated to add the version tag to the other tags if it does not exist already.

However, I'm actually not 100% sure where the version tag came from before. E.g. a bunch of projects would not specify any tags and still get the correct publish.

CRogers (approved):

src/main/groovy/com/palantir/gradle/docker/DockerExtension.groovy
fletch3555: @CRogers, this was a bad change.  Now there's no way NOT to build the version tag (#516)

PR #530, opened 2022-04-04
Implement configuration avoidance

huashenliang:
## Before this PR

The current state of creating tasks is that all of them are being created eagerly.
Therefore, any custom tasks that depend on those docker tasks will be eagerly created.
Readmore on configuration avoidance here: https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:how_does_it_work

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

## Possible downsides?

I don't think there will be a downside, it is mainly syntax changes to use configuration avoidance API.


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @huashenliang! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #560, opened 2022-05-27
fix: handle a blank version

michaelbrewer:
## Before this PR

Handle when the gradle config includes a blank version. Currently this fails with

```
> begin 0, end 1, length 0
```

## After this PR

Following should not raise an exception 

```groovy
plugins {
    id 'com.palantir.docker'
}

version = ""
docker {
    name 'foo'
    tag 'withTaskNameByTag', 'foo:new-latest'
}
```

## Possible downsides?

Blank `version` will now default to `unspecified`

### DIFF
--- a/changelog/@unreleased/pr-560-v1.yaml
+++ b/changelog/@unreleased/pr-560-v1.yaml
@@ -1,0 +1,5 @@
+type: fix
+improvement:
+  description: Allow for version to be blank
+  links:
+    - https://github.com/palantir/gradle-docker/pull/560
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @michaelbrewer! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

Issue #557, opened 2022-05-17
Can't get multiarch buildx to work


ben-skyportsystems:
## What happened?

I'm trying to build a multi-arch image (linux/arm64 and linux/amd64), but can't quite get there.  When I run a "docker" task, it builds everything, but the resulting image goes in the build cache:
```
WARNING: No output specified for docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
```
The `buildx` feature supports a `load` term, but this isn't supported by Docker when multiple architectures are specified:
```
> Task :tools:deployment:controltower:docker FAILED
error: docker exporter does not currently support exporting manifest lists
```
If I try to do run the `dockerPush` task, it fails because it doesn't find an appropriately tagged image.  Not surprising, since the build was build cache-only.
```
> Task :tools:deployment:controltower:dockerTagLatest FAILED
Error response from daemon: No such image: docker.gh.st/repository/ghost/controltower:latest
```
## What did you want to happen?

Running `docker buildx build` manually, I always have to use a `--push` switch.  Maybe that's needed for this plugin until Docker supports locally storing multi-arch images?

Or am I missing something?

## Environment details
* Apple Macbook pro with M1 Max
* MacOS Monterey 13.3.1
* Docker 20.10.14 build a224086
* Gradle 7.4.1
* Plugin 0.33.0

yiweny:
Hi @ben-skyportsystems. 
The issue is resolved in this PR: https://github.com/palantir/gradle-docker/pull/571
It's also mentioned in the official docs of docker that loading to local registry does not work with multi-platform builds. 

Currently, multi-platform images cannot be exported with the docker export type. The most common usecase for multi-platform images is to directly push to a [registry](https://docs.docker.com/engine/reference/commandline/buildx_build/#registry) (see registry)

https://docs.docker.com/engine/reference/commandline/buildx_build/

PR #571, opened 2022-06-21
add --push flag to buildx

yiweny:
## Before this PR
The buildx feature of gradle-docker plugin doesn't support `--push`

## After this PR
`--push` flag supported for the gradle-docker plugin
This would solve the issue: https://github.com/palantir/gradle-docker/issues/557

### DIFF
--- a/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
@@ -181,6 +181,9 @@ class PalantirDockerPlugin implements Plugin<Project> {
             if (ext.load) {
                 buildCommandLine.add '--load'
             }
+	    if (ext.push) {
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/palantir/gradle/docker/PalantirDockerPlugin.groovy
```diff
@@ -181,6 +181,9 @@ class PalantirDockerPlugin implements Plugin<Project> {
             if (ext.load) {
                 buildCommandLine.add '--load'
             }
+	    if (ext.push) {
+	        buildCommandLine.add '--push'
```
bjlaub, line 185: note that `docker buildx build` does not allow `--load` and `--push` to be combined. Might be worth adding something like:
```
if (ext.load) {
    throw new RuntimeError("cannot combine 'push' and 'load' options");
}
```
to make it clear.

fawind (approved):

PR #445, opened 2021-05-06
Add 'quiet' parameter to reduce noisy output

saschpe:
Passes '--quiet' to docker build in order to reduce the build ouput
significantly. 

## Before this PR
The plugin pollutes the Gradle log too much, especially when multiple Gradle modules with containers are part of the same project.

## After this PR
Less noise.

## Possible downsides?
None, the flag is optional and off by default.

PR #572, opened 2022-06-27
Add 'quiet' parameter to reduce noisy output

saschpe:
Passes '--quiet' to docker build in order to reduce the build ouput
significantly. 

## Before this PR
The plugin pollutes the Gradle log too much, especially when multiple Gradle modules with containers are part of the same project.

## After this PR
Less noise.

## Possible downsides?
None, the flag is optional and off by default.

The first pull request went stale due to a lack of reviews (https://github.com/palantir/gradle-docker/pull/445)

### DIFF
--- a/changelog/@unreleased/pr-572.v1.yaml
+++ b/changelog/@unreleased/pr-572.v1.yaml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add 'quiet' parameter to reduce noisy output
+  links:
+    - https://github.com/palantir/gradle-docker/pull/572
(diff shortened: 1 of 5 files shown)


saschpe:
@fawind This has been hanging around for quite some time, maybe you could have a look? TIA.

changelog/@unreleased/pr-572.v1.yaml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add 'quiet' parameter to reduce noisy output
+  links:
+    - https://github.com/palantir/gradle-docker/pull/572
```
saschpe, line 5: It needs a bit of guesswork to get the link right here ;-)

saschpe:
Looks like not much is gonna happen here.

PR #582, opened 2022-08-04
Add platform flag for docker build without a reliance on docker buildx

pguinard-public-com:
## Before this PR

Platform flag is only supported for docker buildx and not docker.

Example of the docker command we would like to run:

docker build --platform linux/arm64 -t example .

## After this PR
The platform flag will be supported such as this to build for a single platform:

```
docker {
    name "example:$version"
    platform "arm64"
```

## Possible downsides?

On upgrade from one version to a version with this commit in it while having the platform flag in a non working state would cause gradle to throw errors rather than silently ignore.


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @pguinard-public-com! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #605, opened 2022-11-22
Support target parameter for multi-stage Docker builds

johndemic:
## Before this PR
The plugin doesn't support the "target" parameter for "docker build"

## After this PR
Added an optional "target" parameter that takes a single String specifying the target parameter for docker build

## Possible downsides?
Don't think are any


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @johndemic! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

Issue #618, opened 2023-02-13
Gradle 8.0 Support


tomasAlabes:
## What happened?

```shell
./gradlew docker

FAILURE: Build failed with an exception.

* Where:
Build file '/home/user/myApp/build.gradle.kts' line: 7

* What went wrong:
An exception occurred applying plugin request [id: 'com.palantir.docker', version: '0.34.0']
> Failed to apply plugin 'com.palantir.docker'.
   > Could not find matching constructor for: org.gradle.api.internal.artifacts.publish.ArchivePublishArtifact(org.gradle.api.tasks.bundling.Zip_Decorated)

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org
```

## What did you want to happen?

Normal task excution

ikstewa:
I tried to take a first pass at upgrading this to Gradle 8 but it's going to be quite an extensive upgrade. This plugin uses a handful of palantir convention plugins that will need to be upgraded as well.

Best approach is to see if we can remove the dependency on internal classes and support usage in other projects which are gradle 8.

sanggggg:
upgrading plugin version to 0.35.0 will resolve this issue 😄

natl-set:
So this is done?

maguro:
For Gradle 8.4, I am getting

    An exception occurred applying plugin request [id: 'com.palantir.docker', version: '0.35.0']
    > Failed to apply plugin 'com.palantir.docker'.
       > Could not find matching constructor for: 
    org.gradle.api.internal.artifacts.publish.ArchivePublishArtifact(org.gradle.api.tasks.bundling.Zip_Decorated)

On my laptop:
```
$ gradle -version
executing gradlew instead of gradle

------------------------------------------------------------
Gradle 8.4
------------------------------------------------------------

Build time:   2023-10-04 20:52:13 UTC
Revision:     e9251e572c9bd1d01e503a0dfdf43aedaeecdc3f

Kotlin:       1.9.10
Groovy:       3.0.17
Ant:          Apache Ant(TM) version 1.10.13 compiled on January 4 2023
JVM:          11.0.17 (Azul Systems, Inc. 11.0.17+8-LTS)
OS:           Mac OS X 14.0 x86_64
```

PR #638, opened 2023-04-03
do not use gradle internal ArchivePublishArtifact

sanggggg:
## Before this PR
Plugin is not compatible with gradle version upper 8.0 #618 

## After this PR
Gradle plugin can't use internal class `ArchivePublishArtifact` anymore upper 8.0. https://github.com/JetBrains/gradle-intellij-plugin/issues/1272
Do not use internal class `ArchivePublishArtifact` directly, instead make artifact with recommended API.


palantirtech:
Thanks for your interest in [palantir/gradle-docker](https://github.com/palantir/gradle-docker), @sanggggg! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

fawind (approved):
Thanks for the fix!

PR #862, opened 2025-04-25
More readme detail about EOL

CRogers:
## Before this PR
Readme suggests this repo will continue to be kept working; this is not the case - we do not use this internally any more.

## After this PR
Put a more realistic description of it's current state.


felixdesouza (approved):

ash211:
Is this superseded by the internal `gradle-sls-docker` plugin these days?

